PR #46, opened 2012-08-17
missing initial scale property

senlin:
add initial scale property to viewport meta tag


iandunn:
Just for the historical record, `initial-scale=1` was added in 129c2382392e3d5253d02e8baefb5344b946e3ee / #285.

senlin:
Not with the original attribution, but glad to know it was added eventually. Thanks for updating it for the record.

PR #460, opened 2014-04-14
Move $content_width within setup function

grappler:
This was raised by @chipbennett in a few ticket on WordPress.org

### DIFF
--- a/functions.php
+++ b/functions.php
@@ -5,13 +5,6 @@
  * @package _s
  */
 
-/**
- * Set the content width based on the theme's design and stylesheet.
- */
-if ( ! isset( $content_width ) ) {
-	$content_width = 640; /* pixels */
-}
-
 if ( ! function_exists( '_s_setup' ) ) :
 /**
  * Sets up theme defaults and registers support for various WordPress features.
@@ -64,6 +57,18 @@ function _s_setup() {
 }
 endif; // _s_setup
 add_action( 'after_setup_theme', '_s_setup' );
+
+/**
+ * Set the content width in pixels, based on the theme's design and stylesheet.
+ *
+ * Priority 0 to make it available to lower priority callbacks.
+ *
+ * @global int $content_width
+ */
+function _s_content_width() {
+	$GLOBALS['content_width'] = apply_filters( '_s_content_width', 640 );
+}
+add_action( 'after_setup_theme', '_s_content_width', 0 );
 
 /**
  * Register widget area.


philiparthurmoore:
We're coming up on a year since this issue was raised...

Anyone?

obenland:
There is a good chance plugins might add callbacks on `after_setup_theme` with a priority of 10 or higher. For them `$content_width` would not be set.

We could debate about whether we should bump the priority of the setup callback to circumvent that, or just leave it as is and stay maximally backwards compatible.

chipbennett:
I'll ask again: any real-world examples of a Plugin breaking?

obenland:
> I'll ask again: any real-world examples of a Plugin breaking?
Not to my knowledge, but I'm sure you'd agree that chance are high that plugins would hook into `after_setup_theme` with the default priority, expecting `$content_width` to be set.

I don't feel particularly strong about this issue either way. I'd consider moving it in the callback and changing its priority to 0 reasonable.

philiparthurmoore:
> I don't feel particularly strong about this issue either way. I'd consider moving it in the callback and changing its priority to 0 reasonable.

I also don't feel as strongly about this issue as I do about getting it resolved. This isn't a nuclear reactor we're handling. If something breaks then so be it; we can revert the change. I like @obenland's idea. @chipbennett or @grappler would you be willing to submit an updated PR for this? Thank you.

grappler:
The additional change that is needed is: `add_action( 'after_setup_theme', '_s_setup', 0 );`? Correct?

fklein-lu:
I think the theme setup function should be kept to the default priority, so that the function is easier to unhook for beginners. Having to deal with different priorities can be difficult.

Let's move the content width into a dedicated function, which is then hooked to `0`. This function could then handle all the content widths, for example if different widths are need for different templates.

On a side note having a smaller theme setup function is also beneficial because I have created a few child themes recently and having to copy a gigantic setup function only to override a single item is kind of a pain.

philiparthurmoore:
I agree with @fklein-lu here.

grappler:
I like @fklein-lu idea. So what do you think of:
```php
/**
 * Set the content width based on the theme's design and stylesheet.
 */
function _s_set_content_width() {
	if ( ! isset( $content_width ) ) {
		$content_width = 640; /* pixels */
	}
}
add_action( 'after_setup_theme', '_s_set_content_width', 0 );
```

fklein-lu:
Three things:
1. We need to declare the `$content_width` as a global inside the function. Else the variable will not be set outside of the function's scope.
2. I don't think we need the `isset()` if we go the dedicated function route, since overriding the variable's value can be achieved in other ways.
3. Would it make sense to make the function pluggable?

chipbennett:
Agreed on 1 and 2. As a counter suggestion, would it make more sense to add
a filter instead of making the callback pluggable?

```
$content_width = apply_filters( '_s_content_width', 640 );
```

grappler:
I agree. Latest version:
```php
/**
 * Set the content width based on the theme's design and stylesheet.
 */
function _s_set_content_width() {
	global $content_width;
	$content_width = apply_filters( '_s_content_width', 640 ); /* pixels */
}
add_action( 'after_setup_theme', '_s_set_content_width', 0 );
```

obenland:
How about:

```php
/**
 * Set the content width in pixels, based on the theme's design and stylesheet.
 *
 * Priority 0 to make it available to lower priority callbacks.
 *
 * @global int $content_width
 */
function _s_content_width() {
    $GLOBALS['content_width'] = apply_filters( '_s_content_width', 640 );
}
add_action( 'after_setup_theme', '_s_content_width', 0 );
```

The comment probably needs a bit more <3.

obenland:
Gah, no emoji support!

chipbennett:
I have no preference concerning `global $content_width` vs.
`$GLOBALS['content_width']`.

philiparthurmoore:
Noting that I'll be reviewing this and getting it into _s next week. If there is any additional feedback speak now. Otherwise, it's time we buried this issue.

glueckpress:
@obenland … except [$GLOBALS['content_width']] will frickin’ fail the theme check. :P 
Theme Check plugin currently looks for `$content_width`explicitly. This will work:

```php
function _s_content_width() {
	$content_width = 640;
	$GLOBALS['content_width'] = apply_filters( '_s_content_width', $content_width );
}
add_action( 'after_setup_theme', '_s_content_width', 0 );
```

I think your comment for the function sounded alright, and I’d love to see this in _s happening soon.

chipbennett:
The @Otto42/theme-check Theme Check Plugin can be fixed, if using $GLOBALS
causes a false-positive. I'd rather fix the algorithm than add in
workarounds. Just need to ping @Otto42.

obenland:
+1

Also, let's keep our language clean here.

glueckpress:
@chipbennett +1.
@obenland Apologies.

Any chance [image sizes](https://gist.github.com/glueckpress/61862e1f30c865b31715) could be considered, or would that be out of scope?

fklein-lu:
The Theme Check plugin for WordPress.org and the VIP Scanner for WordPress.com are on Github, so PRs can be submitted to make these plugins work with the new code.

@glueckpress's latest proposition looks good to me, I'm in favor of merging.

For images, creating a new issue would be the best approach.

philiparthurmoore:
Thank goodness.

mtomas7:
What is the right way to implement additional content width options, if I was using it before like this (Thank you!):

```
if ( ! isset( $content_width ) ) {
	$content_width = 625;
}

if ( ! function_exists( 'tinyframework_content_width' ) ) :
/**
 * Adjust content width in certain contexts.
 *
 * Adjust content_width value for full-width and single image attachment templates, and when there are no active widgets in the sidebar.
 */
function tinyframework_content_width() {
	if ( is_page_template( 'page-templates/full-width.php' ) || is_attachment() || ! is_active_sidebar( 'sidebar-1' ) ) {
		global $content_width;
		$content_width = 960;
	}
}
endif;
add_action( 'template_redirect', 'tinyframework_content_width' );
```

mtomas7:
@fklein-lu May I bug you for assistance with my question above? Thank you!

fklein-lu:
@mtomas7 Please understand that we cannot provide support for using _s.

mtomas7:
I understand. But because this change might affect many users this answer would help them in adapting the theme.

Thank you!

fklein-lu:
You are misunderstanding the implications of this change.

But here's my assistance: open the `index.php` file in the root of your local WordPress install and start reading. Read through every included file until you get to the point where the `template_redirect` action is fired.

Look for:
- When the `functions.php` files are loaded.
- When `after_setup_theme` is fired.
- When `template_redirect` is fired.

That will teach you everything you need to know about this.

mtomas7:
I would like to thank @fklein-lu for the patience and time he spent helping me!

Here is the code for those who want to adjust theme width for other circumstances, including Fränk's comments:

First part is straight from _s `functions.php` file:

```
/**
 * Set the content width in pixels, based on the theme's design and stylesheet.
 *
 * Priority 0 to make it available to lower priority callbacks.
 *
 * @global int $content_width
 */
function _s_content_width() {
	$GLOBALS['content_width'] = apply_filters( '_s_content_width', 640 );
}
add_action( 'after_setup_theme', '_s_content_width', 0 );
```

Then modification of `$content_width`:

```
if ( ! function_exists( '_s_mod_content_width' ) ) :
/* Adjust content width in certain contexts.
 *
 * Adjust content_width value for full-width and single image attachment templates, and when there are no active widgets in the sidebar.
 *
 * template_redirect fires a lot later than after_setup_theme so that will still change the variable as usual.
 *
 * $GLOBALS['content_width'] = 960;
 *
 * is the same as:
 *
 * global $content_width;
 * $content_width = 960;
 *
 */
function _s_mod_content_width() {
	if ( is_page_template( 'page-templates/full-width.php' ) || is_attachment() || ! is_active_sidebar( 'sidebar-1' ) ) {
		$GLOBALS['content_width'] = 960;
	}
}
endif;
add_action( 'template_redirect', '_s_mod_content_width' );
```

:)

mtomas7:
FYI currently this fails Theme Check:

https://wordpress.slack.com/archives/themereview/p1433198197005273

mtomas7:
Looks like it should be all good now :)

Issue #523, opened 2014-06-24
Dropdown menus aren't tabable


chrisdc:
Hello, I was wondering what people's thoughts were on this issue: At the moment you can tab through a menu in _s, but the dropdowns don't become visible. This has been solved in Twenty Fourteen using [some jQuery](https://github.com/WordPress/WordPress/blob/master/wp-content/themes/twentyfourteen/js/functions.js#L106-110), and I'm happy to make a patch based on this if it would be useful, or would a vanilla javascript solution be preferred?

PR #540, opened 2014-07-01
Make dropdown menus accessible by tabbing

chrisdc:
Hello, following on from issue #523 I'm submitting a possible javascript solution to making the dropdown menus in _s tabable (currently dropdowns aren't visible when you tab through the child links). As it's quite a big addition I've broken it down below:

### navigation.js

My additions to navigation.js attempt to replicate [this code](https://github.com/WordPress/WordPress/blob/master/wp-content/themes/twentyfourteen/js/functions.js#L106-110) from Twenty Fourteen, only without the use of jQuery. When a link in a menu is focussed the class `.focus` is added to the ancestors of the link, and when blurred it's removed again. This can be targeted with css to keep the appropriate menus visible. Thanks to @joshmcrty for input on combining this with the existing code.

 **Note** I did include one slight difference from the Twenty Fourteen code: The original adds ` .focus` to every ancestor all the way up to the HTML tag, whereas I have included checks to only include li elements within the current menu systems.

### style.css

To every `:hover` selector in the menu section I've added an equivalent `.focus` selector, reflecting the fact that we want tabbing through menus to be visually as close as possible to using the mouse.

## Limitations

Currently this code uses the variable `menu` previously defined in the navigation toggle code. This is efficient, but does make it relativley complicated for theme authors to add tabbing support to multiple menus. If desired I can adjust this code to work from a separate array of menu IDs.


davidakennedy:
I'm just doubling back to this to say:

1. I've tested the latest and it works well from an accessibility standpoint.
2. We may want to prefix the function names with `_s` like we do with everything else.
3. I'm tagging for a second opinion so we can hopefully get this merged in.

@chrisdc Thanks for all your work on this. Hold off adjusting the function names until we hear a yes or no. :smile:

mor10:
Just bumping: What is the status on this? Can we expect _s to ship with an accessible menu soon?

chrisdc:
The code works well. I've tried to combine flexibility with ease of
modification (allowing for people with limited js skills as far as
possible), but if people don't like the object orientated approach I could
go back to the simpler code I had before. In any case I'm still keen to get
this in _s in some form.

obenland:
> if people don't like the object orientated approach I could go back to the simpler code I had before

I think I'd prefer that. The purpose of this ticket is to improve menu a11y so let's not make it optional. [This is pretty close](https://github.com/chrisdc/_s/blob/74f1149b20954656e5932311344c2cb4505581d3/js/navigation.js) to what it probably should look like eventually.

Also: We as maintainers need to do a better job in shepherding tickets like this, where there is an active and responsive contributor and lots of interest around it. Me explicitly included. Let's make this happen.

chrisdc:
To be honest I'm inclined to agree, I'm starting to feel the oo version might be over-thinking things for a starter theme, so I've switched back for now. I'll leave [this gist](https://gist.github.com/chrisdc/4b09f80281217854622c) up for anyone who's interested in the oo version. The latest code keeps the more compact toggleFocus function and the code to add aria-haspopup.

philiparthurmoore:
> Also: We as maintainers need to do a better job in shepherding tickets like this, where there is an active and responsive contributor and lots of interest around it. Me explicitly included. Let's make this happen.

Completely agree with you here. Where are we at with this? Is it ready for review?

chrisdc:
Does anyone have any thoughts on the issue @davidakennedy raised above (Do we need to prefix js function names)? Other than that it's all ready as far as I can see.

obenland:
Can we use `addEventListener()` and move `toggleFocus()` in there as a callback?

chrisdc:
Do you mean as an anonymous function? Right now I would end up duplicating
code (for onfocus and onblur), but I could split toggleFocus into 2
functions.

PR #579, opened 2014-08-17
Mobile-first menu

sixhours:
Take a mobile-first approach to the main menu toggle, reversing media queries such that the default is the toggle menu, which disappears for larger screens. Also, change pixel values to ems.

### DIFF
--- a/sass/navigation/_menus.scss
+++ b/sass/navigation/_menus.scss
@@ -71,7 +71,7 @@
 	display: none;
 }
 
-@media screen and (min-width: 600px) {
+@media screen and (min-width: 37.5em) {
 
 	.menu-toggle {
 		display: none;
--- a/sass/navigation/_menus.scss
+++ b/sass/navigation/_menus.scss
@@ -62,19 +62,24 @@
 }
 
 /* Small menu. */
-.menu-toggle {
+.menu-toggle,
+.main-navigation.toggled .nav-menu {
+	display: block;
+}
+
+.main-navigation ul {
 	display: none;
 }
 
-@media screen and (max-width: 600px) {
-	.menu-toggle,
-	.main-navigation.toggled .nav-menu {
-		display: block;
-	}
+@media screen and (min-width: 600px) {
 
-	.main-navigation ul {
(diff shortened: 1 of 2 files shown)


iandunn:
I rebased the branch to merge in master and squash the commits, so it's ready to pull again.

In my opinion, though, switching to ems should be done in a separate PR, with all the other media queries included. If it's done here, then we wind up with an inconsistent state where some queries use px and some use ems, and we also have a commit that does two completely unrelated things, rather than being [atomic](http://www.freshconsulting.com/atomic-commits/).

iandunn:
Doh, I just realized that the menu is the only media query in _s, so there wouldn't be any consistency issues, just the atomicity of the commit.

In hindsight, it was probably a bad idea to squash the commits, since they are unrelated, but it's probably fine to have them both on this PR, since both issues were discussed.

It was also bad to rebase at this stage, since it deletes the original commits (and any comments attached to them) and creates new ones. Ideally rebasing should only be done during the merge, so that it never affects the original branch. That was my bad, sorry about that :(

From what I remember, the comments that were lost were @obenland giving the thumbs up for the switch to ems, and linking to http://blog.cloudfour.com/the-ems-have-it-proportional-media-queries-ftw/ as justification. Then, someone else brought up a Chrome bug related to using ems, but it didn't seem to be a blocker.

Since the damage is already done, I rebased again to re-separate the commits.

davidakennedy:
Tested, and looks good. Suggesting merge.

obenland:
Do it!

slushman:
In the interest of being DRY, would be better to move lines 70-72 up to where main-navigation ul is already (line 11) rather than redeclaring it again down with the media query?

obenland:
Yes it would, good catch.

philiparthurmoore:
Has anyone tested this? I see no label for needs testing and there is a recent commit. Happy to test this if it's ready.

obenland:
Haha @sixhours, you're my puppet now!

philiparthurmoore:
:puppet:

Issue #543, opened 2014-07-18
Bug: skip-link-focus-fix.js causes unexpected outline on elements in Chrome


SimonEast:
**skip-link-focus-fix.js** is supposed to correct the tab order for keyboard navigation to increase usability.  One unintended side-effect, however is that once the script adds `tabindex="-1"` to elements on `hashchange`, that DIV can now receive focus PLUS it also receives Chrome's `:focus` styling which is a blue outline.  After the hash-change event, you can simply click on the DIV and an outline appears in blue.  That's not too pretty.

**Steps to recreate issue in Chrome:**

1. Load a URL that contains skip-link-focus-fix.js
2. Pick a DIV on the page that has an ID
3. Append `#id-goes-here` to the URL
4. That element now has a blue border, and anytime you click on that element the blue border returns.

**Workaround**
You can apply some CSS overrides such as:
```
:focus {
    outline: none;
}
```
or
```
#myElement:focus {
    outline: none;
}
```

davidakennedy:
Doubling back to this since it's got a bit stale.

Let's get this in. @jefferyto makes a solid argument, and even though the "skip link bug" is getting fixed in browsers, it will likely be awhile before it lands in Safari and Chrome.

@jefferyto Can you clean up your patch, sync with master and make an official pull request.

Also, in running the possible solution through JS Hint, I get two warnings we may need to address:

```bash
14	Use '!==' to compare with 'null'.
26	Use '===' to compare with '0'.
```

Thanks!

davidyezsetz:
@davidakennedy I've wrapped this up in https://github.com/Automattic/_s/pull/755

karmatosed:
@davidakennedy any ideas on this as I see you are active around it.

davidakennedy:
@karmatosed The latest is here:

https://github.com/Automattic/_s/pull/814

Lots of feedback and thoughts, trying to move forward with it.

lachlancanard739:
> hi [@davidakennedy](https://github.com/davidakennedy) if you do the same test changing the URL adding #post-1 you wiil see the outline on the article element. That's because CSS reset takes care of the outline only for a subset of elements, including DIVs (and #comments is a div) while #post-1 is an article. Teorically, this would happen for all the elements not included in the outline CSS reset like for example: article, aside, canvas, details, embed, figure, figcaption, footer, header, hgroup, menu, meter, nav, output, ruby, rb, rt, rp, section, summary... etc.
> 
[...]

PR #604, opened 2014-08-31
Improvements for skip-link-focus-fix.js

jefferyto:
These are some (somewhat slight) improvements that came up while I was looking at #543:

* fixed browser detect for IE11
* add `tabindex` only if the element doesn't have one already
* remove `tabindex` we added once the user has tabbed off

Would be nice to have this merged before working on a proper PR for #543.


karmatosed:
Hey @davidakennedy - if you've got a chance, taking a look at this would be great.

Issue #619, opened 2014-09-23
Organized template parts - a proposal


mayeenulislam:
- `content-none.php`
- `content-page.php`
- `content-search.php`
- `content-single.php`
- `content.php`

a whole lot of files, I think, to be re-arranged somewhere. If we can introduce a new folder structure with `/template-parts` and organize all these parts there, isn't that good looking and more organized?

If so then we just have to call all the `get_template_part()` like:
`get_template_part( 'template-parts/content', , get_post_format() );`

PR #642, opened 2014-11-04
New "template-parts" folder & Jetpack Infinite scroll

leopuleo:
New file template organization + Jetpack Infinite Scroll support by @grappler.
This PR is referring to leopuleo #637, now closed, and discussion #619.


philiparthurmoore:
I'm shocked this hasn't been merged yet. Sorry for dropping the ball on this conversation @leopuleo. This is still a very good pull request that I would love to get merged in now. Can you please review everything and let me know if it's ready for testing? I'll be happy to give this a fresh look.

philiparthurmoore:
Fixed in https://github.com/Automattic/_s/commit/71d4231cbebb44800739cb50abac8b251977676f.

mtomas7:
@leopuleo @philiparthurmoore I noticed that in order to accommodate this file structure, a special approach was needed to make Jetpack feature to work. Isn't it a weak side, because plugins by default would assume that content-* files are in the main WP dir?

Is it possible that there can be much greater number of plugins that would not work correctly with such theme?

Thank you!

philiparthurmoore:
Hi @mtomas7.

> I noticed that in order to accommodate this file structure, a special approach was needed to make Jetpack feature to work.

I don't know that I would call the approach special. It's just another way of rendering the loop that deviates from the way that [Jetpack does it by default](https://github.com/Automattic/jetpack/blob/master/modules/infinite-scroll/infinity.php#L1153). 

```php
	/**
	 * Rendering fallback used when themes don't specify their own handler.
	 *
	 * @uses have_posts, the_post, get_template_part, get_post_format
	 * @action infinite_scroll_render
	 * @return string
	 */
	function render() {
		while ( have_posts() ) {
			the_post();

			get_template_part( 'content', get_post_format() );
		}
	}
```

> Isn't it a weak side, because plugins by default would assume that content-* files are in the main WP dir?

I don't think that any plugin should make this assumption. The way that IS is implemented in _s doesn't really relate to how plugins interact with the theme. As long as we make sure that the IS render loop is properly defined in the theme then everything should be fine. If a plugin has an issue with it then it's probably a mistake in the plugin which I'd be happy to look into and help you out with.

Cheers.

PR #647, opened 2014-11-14
Add missing escaping

nickdaugherty:
Adds a variety of escaping improvements (mostly to translations) to harden the theme against XSS, XSRF, and other badness.

### DIFF
--- a/404.php
+++ b/404.php
@@ -12,11 +12,11 @@ get_header(); ?>
 
 			<section class="error-404 not-found">
 				<header class="page-header">
-					<h1 class="page-title"><?php _e( 'Oops! That page can&rsquo;t be found.', '_s' ); ?></h1>
+					<h1 class="page-title"><?php esc_html_e( 'Oops! That page can&rsquo;t be found.', '_s' ); ?></h1>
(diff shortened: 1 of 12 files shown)


philiparthurmoore:
Based on the outcome of https://core.trac.wordpress.org/ticket/30724 I'm going to close this ticket. If there's a future proposal for better escaping in _s then please feel free to open another PR; for now though I don't see the value of keeping this ticket open.

nickdaugherty:
The core ticket referenced refers to the Twenty Fifteen theme, whose translations can be closely controlled by a team of experts.

_s is a building block theme...and as such, sites will use their own translations, which will almost certainly have much lower quality control. 

Escaping translations here provides a solid and secure base for any theme building on _s, and sets a good example for the community. 

We'd never explicitly trust data from other sources on output...why do translations get a free pass? They are vulnerable to tampering and bad practices, just like anything else.

fklein-lu:
As mentioned previously, translations are used provided content and should not be trusted.

More importantly, escaping is required according to WordPress.com coding standards (this is across all of WordPress.com).

We'll update the code and merge this directly.

PR #650, opened 2014-11-20
Remove Custom Header and Background Admin Callbacks

fklein-lu:
With [#25571](https://core.trac.wordpress.org/ticket/25571) landing in Core, the Custom Header page under Appearance will now redirect to the corresponding section in the Customizer.

This allows us to remove the callback functions that were used to style the old Custom Header page.

### DIFF
--- a/inc/custom-header.php
+++ b/inc/custom-header.php
@@ -19,8 +19,6 @@
  * Set up the WordPress core custom header feature.
  *
  * @uses _s_header_style()
- * @uses _s_admin_header_style()
- * @uses _s_admin_header_image()
  */
 function _s_custom_header_setup() {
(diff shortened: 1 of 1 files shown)


philiparthurmoore:
Following up on this. When 4.3 is released let's get this in. Shouldn't be much longer now that 4.2 is imminent.

philiparthurmoore:
Update. 4.3 is scheduled for release during August 2015.

philiparthurmoore:
Should we go ahead with this now that 4.3 is imminent?

obenland:
Sure.

sarahmonster:
Since this is roughly a million years old and out of date but should definitely be merged, I've updated it as #879.

PR #651, opened 2014-11-25
New heading structure

mor10:
Aims to create a more hierarchical heading structure. Somewhat based on structure in Twenty Fifteen and inspired by http://blog.rrwd.nl/2014/11/21/html5-headings-in-wordpress-lets-fight/ and https://www.joedolson.com/2014/11/headings-hierarchy-problem/

### DIFF
--- a/archive.php
+++ b/archive.php
@@ -25,7 +25,9 @@ get_header(); ?>
 			<?php while ( have_posts() ) : the_post(); ?>
 
 				<?php
-					/* Include the Post-Format-specific template for the content.
+
+					/*
+					 * Include the Post-Format-specific template for the content.
 					 * If you want to override this in a child theme, then include a file
 					 * called content-___.php (where ___ is the Post Format name) and that will be used instead.
 					 */
--- a/comments.php
+++ b/comments.php
@@ -25,7 +25,7 @@ if ( post_password_required() ) {
 	<?php if ( have_comments() ) : ?>
 		<h2 class="comments-title">
 			<?php
-				printf(
+				printf( // WPCS: XSS OK
 					esc_html( _nx( 'One thought on &ldquo;%2$s&rdquo;', '%1$s thoughts on &ldquo;%2$s&rdquo;', get_comments_number(), 'comments title', '_s' ) ),
 					number_format_i18n( get_comments_number() ),
 					'<span>' . get_the_title() . '</span>'
--- a/comments.php
+++ b/comments.php
@@ -25,7 +25,7 @@ if ( post_password_required() ) {
 	<?php if ( have_comments() ) : ?>
 		<h2 class="comments-title">
 			<?php
-				printf( // WPCS: XSS OK
+				printf( // WPCS: XSS OK.
 					esc_html( _nx( 'One thought on &ldquo;%2$s&rdquo;', '%1$s thoughts on &ldquo;%2$s&rdquo;', get_comments_number(), 'comments title', '_s' ) ),
 					number_format_i18n( get_comments_number() ),
 					'<span>' . get_the_title() . '</span>'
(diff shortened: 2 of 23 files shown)


davidakennedy:
Thanks for the PR, @mor10! I think this needs a tad more work. How about:

* Also changing widget titles to a `h2`.
* Accounting for the situation when a user has set a static front page and another page as a post. In that case the Posts page needs a true `h1`. Check out how it was done in [Twenty Fifteen](https://core.trac.wordpress.org/browser/trunk/src/wp-content/themes/twentyfifteen/index.php).

mor10:
I'll see if I can get it done over the weekend. I'm wondering if widget heading should be h3 to indicate clear hierarchical status. Esp when you have setups like Twenty Fifteen where the widgets are displayed before the content in the HTML hierarchy (which is bad form, but outside the scope of this convo). Thoughts?

obenland:
> Thoughts?

I'm thinking: Let's not do that.

mor10:
Hierarchical heading structure is quite important. The reasoning in [this article] (http://blog.rrwd.nl/2014/11/21/html5-headings-in-wordpress-lets-fight/) on sidebar widgets using h3 is sound. I also seem to remember h3 for widgets used to be fairly commonplace until HTML5 came along and everything was changed to h1. 

If we follow the logic of heading hierarchy then it makes sense for the widgets to always be lower down on the chain than the content. Since content titles are either h1 or h2 the next level down is h3. h2 for widgets therefore seems a bit arbitrary as it puts widget titles on the same level as article titles in index pages.

That said my main concern is heading levels in actual articles and the site as a whole so if there's a lot of pushback on the widgets I'm fine with leaving them alone or setting them at h2.

mor10:
@davidakennedy I've added the fallback for index.php but I'm reluctant to add the sidebar heading structure due to the controversy it may cause. Would probably work better as a separate PR.

mrwweb:
Just found this PR and it makes me SO HAPPY. That's for getting this rolling @mor10.

One thought on the site title markup, I often come back to [this survey](http://webaim.org/projects/screenreadersurvey3/#headings) suggesting that a **majority** of screen reader users prefer **two** `<h1>`s per page: one for the site title and one for the page title. I see why this makes sense as many users will enter a site not from the home page but from a referral link, search page, email, social media post, etc.

Frustratingly, I think that question hasn't been asked in more-recent versions of the survey, but it still makes me wonder if we should at least consider two `<h1>`s per page. It's defensible at worst.

mor10:
I will be submitting a new and more expanded version of this PR shortly
addressing sidebar headings as well. Will look at the issue of two
top-level headings at the same time. It'll likely be a separate PR to
replace this one.

In medias res on my paranoid android

philiparthurmoore:
@mor10, should we close this Issue out then? Happy to leave it open until you submit a better PR but trying to take care of some housekeeping.

mor10:
Yes. I will submit a new PR shortly

In medias res on my paranoid android

davidakennedy:
Hey @mor10, do you still have interest in refreshing this pull request? I'd love to get it in a more final state so we can make a decision.

Thanks!

mor10:
I'll try to get it done today

In medias res on my paranoid android

mor10:
@davidakennedy I've made some minor cleanups and changed the widget headings to H2. With this pull request the site title will display as H1 only on the front page while on other pages whatever the actual content of the page is will display with H1 while the site title displays as a div. 

Note that I preserved the site description as a paragraph even though I have no understanding of why it is a paragraph.

grappler:
@mor10 Now you just need to squash all of the commits into one. https://github.com/ginatrapani/todo.txt-android/wiki/Squash-All-Commits-Related-to-a-Single-Issue-into-a-Single-Commit

mor10:
This is getting too messy. I'm killing this pull request and starting a new clean one.

grappler:
@mor10 You can still clean this up.

mor10:
Too late. It's dead. Git is a horrible tool with terrible documentation. I have no patience for reading through 800 awful tutorials that don't solve the problem. Pushing a new clean pull request is much easier.

Issue #545, opened 2014-07-25
Add ARIA support to menu toggle for improved accessibility


davidakennedy:
I'd like to propose that we add ARIA to the Primary Menu toggle and the associated `ul` that gets expanded or collapsed as a result.

Here is a [good primer on ARIA](http://www.marcozehe.de/2014/03/27/what-is-wai-aria-what-does-it-do-for-me-and-what-not/).

I'd like to add these attributes:
- `aria-controls`
- `aria-label`
- `aria-labeledby`
- `aria-hidden`
- `aria-expanded`

I built a quick [prototype on CodePen](http://codepen.io/davidakennedy/pen/Cjmvi).

I figured a quick demo was better so we can discuss if need be, and it's probably better to wait on a PR until #540 is sorted.

Explaining the prototype:
- I add the ARIA attributes with JavaScript. Since they need to be changed with JS anyway, this makes sense, plus it keeps the markup in our `header.php` a bit cleaner. That makes it easier to remove the toggle if a theme author does not need it.
- There is varying support for ARIA depending on the assisitive technology and browser used, but it doesn't hurt to use any of it. It's valid HTML5 and support will only improve.

Markup:
- `aria-label="Toggle Primary Menu"`: Adds a label to the button. Notice it says, toggle, which is more descriptive of what it does than it's visual label. Screen readers with support will announce this rather than text. Helpful, especially if using an icon as the button.
- `aria-labelledby="menu-toggle"`: Maps the button label of element it's label.
- `aria-controls="menu"`: Maps buttons functionality to the thing it's controlling.
- `aria-expanded="true"`: Or false. Is this thing expanded or not? Screen readers with support will announce this, like: "Toggle Primary Menu button expanded"
- `aria-hidden="true"`: Or false. Is this thing hidden or not?

**One thing I'm still researching**: Whether I need both `aria-hidden` and `aria-expanded`. In this case, it may be better to use just `aria-expanded`.

I'm open to suggestions and improvements, and of course, whether this should be included in `_s`.

I think it makes sense as long as we're providing a basic toggle. This would improve the accessibility of the toggle functionality, and provide a nice example to theme developers.

Issue #654, opened 2014-11-28
Validation Errors & Warnings


emiluzelac:
Hey folks,

Before we close this in right away, please give me a moment or two to explain.

At least once a month someone will bring the ```h1``` problem to the table, but only for SEO reasons right? OK, now let's forget the SEO for a sec.

Few moments ago I was trying to validate an ```_s``` based theme I am working on and validator throws the following messages.

## Warnings

*Consider using the h1 element as a top-level heading only (all h1 elements are treated as top-level headings by many screen readers and other tools).*

This would be accessibility if I am not mistaking and our accessibility gurus can confirm.

```<h1 class="screen-reader-text">Posts navigation</h1>```

Second and very similar:

*Consider using the h1 element as a top-level heading only (all h1 elements are treated as top-level headings by many screen readers and other tools).*

```<h1 class="entry-title"><a href="#" rel="b…```

## Errors

We introduced this while ago:

*The aria-controls attribute must point to an element in the same document.*

```…on class="menu-toggle" aria-controls="menu" aria-expanded="false">Primary Menu</button>```

----

```h1``` is a no brainer, switching them to i.e. ```h2``` will eliminate the warnings, but I have no idea how to get around the aria-controls.

At the same time HTML5 does recommend that we don't follow current aproach, see: http://www.w3.org/TR/2014/WD-html5-20140617/sections.html#headings-and-sections

From what I can see, we did not discuss this particular issue, searched for it, but no luck. If we did and you think that I am being too OCD, feel free to close the ticket :)

davidakennedy:
Hey @emiluzelac, we have an open pull request for a [new heading structure](https://github.com/Automattic/_s/pull/651). We'll see what happens there.

I'll check into the warning on the `aria-control`.

emiluzelac:
Great, feel free to handle the ticket status any way you find appropriate :)

jaspermdegroot:
@davidakennedy - My PR https://github.com/Automattic/_s/pull/657 will fix the ```aria-control``` issue.

karmatosed:
I think I am right with the new PR merged this should be closed. Please reopen if I am wrong though.

PR #657, opened 2014-12-03
Header: Made aria-controls refer to an ID instead of a class

jaspermdegroot:
The ```aria-controls``` attribute that was added to the menu-toggle button in https://github.com/Automattic/_s/commit/e142af3a2720e38179e294181562c22842a68c3e refers to the class of the primary menu. According to the specs it should refer to an ID (http://www.w3.org/TR/wai-aria/states_and_properties#aria-controls).

Ref #545

### DIFF
--- a/header.php
+++ b/header.php
@@ -28,8 +28,8 @@
 		</div><!-- .site-branding -->
 
 		<nav id="site-navigation" class="main-navigation" role="navigation">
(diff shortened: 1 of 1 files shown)


davidakennedy:
Tested this and it works well. Fixes the error noted in #654. Thanks @jaspermdegroot!

jaspermdegroot:
@davidakennedy - Thanks for merging!

PR #662, opened 2014-12-10
Make sure entry titles don't overflow content area

lelandf:
As per the theme unit test data, titles shouldn't overflow the content area. It suggests adding word-wrap: break-word; to support non-breaking text. This pull request basically just adds that recommended code to _s.


philiparthurmoore:
Super-long titles are an edge case that we really shouldn't be accommodating for via `break-word`. This has come up before in https://github.com/Automattic/_s/pull/372#issuecomment-31312780. Cheers.

PR #663, opened 2014-12-11
Fix error with travis

miya0001:
It will fix error that caused by updating php-codesniffer.

Anyway, I thought version of the phpcs should be locked like following.

```
$ composer global require "squizlabs/php_codesniffer=2.0.0"
```

But WordPress rule will be update again in the future and it doesn't have tags.
I couldn't decide lock or not ...

### DIFF
--- a/codesniffer.ruleset.xml
+++ b/codesniffer.ruleset.xml
@@ -8,6 +8,8 @@
 	<!-- Include the WordPress ruleset, with exclusions. -->
 	<rule ref="WordPress">
 		<exclude name="Generic.WhiteSpace.ScopeIndent.Incorrect" />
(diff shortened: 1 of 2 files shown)


fklein-lu:
Thanks for the PR, that Travis error is annoying.

@philiparthurmoore I'm not familiar with Travis, can you have a look at this please?

philiparthurmoore:
Sure. I'll fix this today.

miya0001:
Have you figured out another solution?

philiparthurmoore:
What I mean is that I plan on updating that Travis file to avoid these issues. Happy to use your solution as part of the update. These Travis fails are annoying but not unexpected. Just need to update that config file.

miya0001:
Thanks. :)

miya0001:
:+1:

PR #666, opened 2014-12-17
Fix child menus for touch devices

sixhours:
Add the ability to click on the parent once to access submenus, twice
to access the parent item. Props @iamtakashi for the original fix.

### DIFF
--- a/js/navigation.js
+++ b/js/navigation.js
@@ -41,4 +41,35 @@
 			menu.setAttribute( 'aria-expanded', 'true' );
 		}
 	};
+	
+	// Fix child menus for touch devices.
+	function fixMenuTouchTaps( container ) {
+		var touchStartFn,
+		    parentLink = container.querySelectorAll( '.menu-item-has-children > a, .page_item_has_children > a' );
+
+		if ( 'ontouchstart' in window ) {
+			touchStartFn = function( e ) {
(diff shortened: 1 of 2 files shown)


samikeijonen:
I only had iPad for testing. This works great in native Safari Browser but not in Chrome. I wonder if this is a bug in Chrome or in code.

davidakennedy:
Hey @sixhours, when you get a chance, I'd love to see a refresh on this.

I think addressing some of Obenland's feedback above would be great. Then we can test more thoroughly.

I'm in favor of this getting in though.

sixhours:
> Defining a function and then immediately calling it seems not very elegant. It would also be great for it to have a better documentation than simply stating that it's a fix.

It's a bit over my head, honestly. If someone with more JS-fu wants to take a look at it, that'd be cool.

sixhours:
This works for me in testing! We've also used it in many themes on WordPress.com with success. I'll see if I can resolve the conflicts and merge it.

mtomas7:
@sixhours Did you use code from: https://github.com/Automattic/_s/pull/900 or the one that is attached to this issue?

sixhours:
Oh good catch, I meant to grab from #900. Will create a new PR; closing this one to avoid confusion.

Issue #659, opened 2014-12-03
_s' Audience


philiparthurmoore:
I'm finding it harder to determine which audience to code for and merge pull requests for as _s becomes more popular. There are a few major "stakeholders" as far as _s goes right now:

- WordPress.com / Automattic / Jetpack
- WordPress.org theme developers who download the theme from [underscores.me](http://underscores.me)

Automattic is the granddaddy of _s and the theme team uses _s for all WordPress.com theme development, as well as recommends it to premium theme shops on WordPress.com. This is super-important. When I think about this I want to add stuff into _s that will make it much easier to develop for Jetpack and such, and I also want to adhere to the VIP and WordPress.com scanners as much as possible.

WordPress.org theme developers have taken a very strong liking to _s, which is so nice. The big consideration here is that the theme review team has requirements. When _s is in conflict with those requirements then we are faced with making it difficult for theme developers to make good stuff easily, which is one of the goals of _s. This is when I think about the theme check plugin.

Is there a case to be made for two branches of _s, a .com branch and a .org branch? Might this alleviate some of the issues that we're facing in terms of the final say on pull requests and also make it easier to start adding more things into _s that make sense, like all the Jetpack include files that .org users might not really need?

At the end of the day this is an Automattic project but so many .org devs are using _s and it's getting a bit difficult to figure out how best to handle PRs and such. I just want to blaze through these issues and pull requests so furiously but I'm really scared of upsetting either camp. I'm in the lucky position of feeling empathy for both sides right now but the unlucky position of possibly messing something up. :smile:

PR #671, opened 2014-12-25
Theme Hook Alliance

protechig:
I am working on an independent fork of _s to help my team (bourbon, neat, etc..). One thing that I love about the Genesis Framework is their robust hooks - which lead me to come across Doug Stewart's Theme Hook Alliance. I added them to my theme. I created a separate THA branch that only adds the THA functionality so it could be merged directly to _s if you choose. I saw an issue posted in January about it and the conclusion is that it would be considered. Here it is! This is my first github pull request.

Thanks, and Merry Christmas!

### DIFF
--- a/content-search.php
+++ b/content-search.php
@@ -8,6 +8,7 @@
  */
 ?>
 
+<?php tha_content_top(); ?>
 <article id="post-<?php the_ID(); ?>" <?php post_class(); ?>>
 	<header class="entry-header">
 		<?php the_title( sprintf( '<h1 class="entry-title"><a href="%s" rel="bookmark">', esc_url( get_permalink() ) ), '</a></h1>' ); ?>
@@ -27,3 +28,4 @@
 		<?php _s_entry_footer(); ?>
 	</footer><!-- .entry-footer -->
 </article><!-- #post-## -->
+<?php tha_content_bottom()l ?>
--- a/404.php
+++ b/404.php
@@ -8,8 +8,10 @@
 get_header(); ?>
 
 	<div id="primary" class="content-area">
+    <?php tha_content_top(); ?>
 		<main id="main" class="site-main" role="main">
-
+            <?php tha_entry_before(); ?>
+            <?php tha_entry_top(); ?>
 			<section class="error-404 not-found">
 				<header class="page-header">
 					<h1 class="page-title"><?php _e( 'Oops! That page can&rsquo;t be found.', '_s' ); ?></h1>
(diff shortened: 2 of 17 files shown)


jacklenox:
It's my understanding that the THA is primarily focussed on improving child theming. As @emiluzelac said above `_s` is a starter theme and creating a child of it is specifically discouraged.

I also don't see how these hooks are a good fit.

grappler:
@jacklenox '_s' is a starter theme but all of the themes built off it will be parent themes so it makes sense for the parent theme based on _s to improve child themeing. These hooks can also be used in plugins. There is even a plugin on w.org which uses the hooks.

jacklenox:
@grappler Good point. :)

samikeijonen:
Is there a reason why hooks `tha_entry_before` and `tha_entry_top` are basically in the same spot in `404.php` file?

ianstewart:
I'm not a fan of adding THA hooks to _s. It moves very far away from the leanness of _s and tons of hooks doesn't necessarily make child theming better.

sterndata:
I've been loosely following this thead and I have to agree with ianstewart:  This isn't Genesis. It's lean, mean, quick.

content-search.php
```diff
@@ -27,3 +28,4 @@
 		<?php _s_entry_footer(); ?>
 	</footer><!-- .entry-footer -->
 </article><!-- #post-## -->
+<?php tha_content_bottom()l ?>
```
grappler: There is an error here. The `1` should be `;`

grappler:
This would be an interesting feature as discussed in https://github.com/Automattic/_s/issues/659 People are using `_s` for differnt thing e.g. Custom themes for clients, themes to be distributed to the mass or WordPress.com themes. Perhaps we could add this as an option which is disabled by default. It would make it easier for some of us if we could download `_s` with the hooks.

This is not just for child themes but also having standard hooks which plugins can hook into.

For refernce THA came out of this WP core ticket discussion: https://core.trac.wordpress.org/ticket/21506

content-search.php
protechig: I'll get that fixed!

protechig:
I'm impartial as to whether THA makes sense in `_s` or not. While `_s` isn't Genesis (obviously) one of the features that I really do enjoy about genesis is it's plethora of action hooks. That is why I forked it and added it to my fork - as I want to make a pseduo-framework / parent theme that suits my personal and professional needs. In the nature of GPL, I figured I'd submit a pull request and if enough people found it useful or wanted  then that's great!

I definitely do think `_s` does have multiple audiences as @grappler referenced, and it may be useful to the community to have different branches for different types of themes (.com vs. org) as well as be able to enable different features on download - why couldn't THA be a checkbox just like Sass is?

obenland:
I agree with @emiluzelac, @jacklenox, and @ianstewart. It's just not a good fit at the moment, even as an optional feature.

Let's revisit this once there is a core supported solutions in place.

Thanks for your work on this PR @protechig!

Issue #617, opened 2014-09-14
Introduce normalize.css into Underscores


philiparthurmoore:
This has been brought up previously: #3, #44, #174, #267.

In general I am 100% in favor of ditching a hard reset and other random bits and moving to normalize, but we'll obviously need a little bit of discussion around this. /cc @ianstewart, @obenland, @karmatosed.

obenland:
Fixed in ea84ceec956152db7ec5f221eaf9dcab283a3751.

Issue #174, opened 2013-03-05
Better to use normalize instead of a reset


webvitalii:
There is no need to reset all browser styles.
It is much more easier to extend the default styles with normalize.css
http://necolas.github.com/normalize.css/
http://stackoverflow.com/questions/6887336/what-is-the-difference-between-normalize-css-and-reset-css

Issue #44, opened 2012-08-16
Use Normalize.css instead of hard reset


nvartolomei:
%subj%

PR #674, opened 2014-12-31
Normalize rather than Reset.

obenland:
The reset we used was a mix of Normalize, the Paul Irish reset,
sprinkled with some blueprint.css, and base styles. It was hard to
maintain and just seemed outdated (setting the base font-size to 10px
instead of 16px, among other things).

Normalize sets saner defaults and is generally just not as disruptive
as a full reset.

The code was added as is, has only its comments stripped (for now), and
is not mixed with any styles to make future updates to it as easy as
possible.

See #3, #44, #174, #267, #617.


fklein-lu:
:+1:+ PR is looking great.

lancewillett:
+1

PR #677, opened 2015-01-09
Reflect the change of color in the customizer.

mignonstyle:
Reflect the change of color of the title in the customizer.
Fixed ".site-title" to ".site-title a".

If you change the header text color in the customizer, it will not be reflected immediately.
By changing the ".site-title" of customizer.js to ".site-title a", the header text color will now be reflected immediately.

### DIFF
--- a/js/customizer.js
+++ b/js/customizer.js
@@ -20,12 +20,12 @@
 	wp.customize( 'header_textcolor', function( value ) {
 		value.bind( function( to ) {
 			if ( 'blank' === to ) {
-				$( '.site-title, .site-description' ).css( {
+				$( '.site-title a, .site-description' ).css( {
 					'clip': 'rect(1px, 1px, 1px, 1px)',
 					'position': 'absolute'
 				} );
 			} else {
-				$( '.site-title, .site-description' ).css( {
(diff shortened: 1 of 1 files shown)


timnicholson:
I was going to post a new issue here, but found this one. This fix is nice, but still doesn't work completely right. What it does is always set the site-title link to the default text color. Except that is not at all how the site-title is displayed on the front end. The site-title link color is always set in CSS. 

The Javascript would need to also know what the default text color is and if the current color is the default, then I guess set it to color: inherit. Right now even with this fix, it always sets it to the default text color (from the add_theme_support) not the actual default link color (from style.css).

The custom-header.php file knows what the default color is, but since customizer.js is "static" (meaning not generated by PHP), there's no way to easily pass it the default link color.

I guess what would be best is if the add_theme_support had a new parameter for default-link-color. Then customize.js would need to be generated by PHP or inline'd. Or just inline setting the link color at least.

karmatosed:
I think as this fix partially fixes it, lets pull this in. If you can also think of a way to improve this please send in a further PR.

mignonstyle:
Thank you. I will think of a way to further improve.

PR #679, opened 2015-01-11
Remove unnecessary margin-top and margin-bottom from center-block mixin.

iandunn:
Setting top and bottom margins is not necessary to center the element, and it can potentially override the margins set by another rule.


ScottSmith95:
Nice change! Looks good.

PR #680, opened 2015-01-11
Set a fixed table-layout in the clearfix to avoid max-width conflict

iandunn:
I'm not sure if this is a problem or not, but I figured I'd open a discussion to document it and see what other people think.

Setting `display: table` in the clearfix can lead to some browsers ignoring the max-width property on child elements (see links at the bottom). I've created [a minimal example](https://iandunn.name/content/misc/2015/_s-clearfix-max-width.html), so you can see the issue by opening it in the Firefox 34 or IE 11.

I'm hesitent to say this is an actual issue, since the CSSMojo clearfix is pretty popular and I wasn't able to find any other discussions about the conflict with it specifically, but at least in my limited experience, it seems like it's a problem.

So far, adding `table-layout: fixed` to the `clearfix` mixin solves the problem for me with no side-effects.

### References
* http://www.carsonshold.com/2014/07/css-display-table-cell-child-width-bug-in-firefox-and-ie/
* http://stackoverflow.com/a/11310261/450127
* https://bugzilla.mozilla.org/show_bug.cgi?id=975632

### DIFF
--- a/sass/mixins/_mixins-master.scss
+++ b/sass/mixins/_mixins-master.scss
@@ -14,6 +14,7 @@
 @mixin clearfix() {
 	content: "";
 	display: table;
+	table-layout: fixed;
 }
 
 // Clear after (not all clearfix need this also)
--- a/style.css
+++ b/style.css
@@ -599,6 +599,7 @@ a:active {
 .site-footer:after {
 	content: "";
 	display: table;
+	table-layout: fixed;
 }
 
 .clear:after,


philiparthurmoore:
@iandunn: Sorry to bother you with this but would you mind putting some screenshots here so we can see what's up, if even for posterity? I want to make sure I'm seeing the right thing.

iandunn:
No problem, 

This screenshot shows the example without any modifications, and you can see that the `max-width` setting on the image is not being applied:

max-width-not-applied-because-display-table

This screenshot has `table-layout: fixed` added to the clearfix, which fixes the issue so that `max-width` applies to the image:

max-width-applied-because-table-layout-fixed

philiparthurmoore:
@iandunn Thank you very much for this. I've run into similar issues with tables overflowing their containers and needing `table-layout: fixed`. I'll play with this a bit and merge it in if everything looks good.

karmatosed:
@philiparthurmoore what are your thoughts after playing? I see you say you're going to play a bit no this?

PR #682, opened 2015-01-18
Lay out content and widget areas with inline-block instead of floats.

iandunn:
This is the current best practice, and it avoids the need for the clearfix hack because inline-block elements aren't pulled out of the document flow.

### DIFF
--- a/layouts/content-sidebar.css
+++ b/layouts/content-sidebar.css
@@ -5,7 +5,7 @@ Layout: Content-Sidebar
 
 .content-area {
 	display: inline-block;
-	margin: 0 -25% 0 -6px;
+	margin: 0 -25% 0 -6px;	/* left margin offsets the inline-block spacing, see http://css-tricks.com/fighting-the-space-between-inline-block-elements/ */
 	padding-left: 6px;
 	width: 100%;
 }
--- a/layouts/content-sidebar.css
+++ b/layouts/content-sidebar.css
@@ -4,15 +4,16 @@ Layout: Content-Sidebar
 */
 
 .content-area {
-	float: left;
-	margin: 0 -25% 0 0;
+	display: inline-block;
+	margin: 0 -25% 0 -6px;
+	padding-left: 6px;
 	width: 100%;
 }
 .site-main {
 	margin: 0 25% 0 0;
 }
 .site-content .widget-area {
-	float: right;
+	display: inline-block;
 	overflow: hidden;
 	width: 25%;
 }
--- a/layouts/sidebar-content.css
+++ b/layouts/sidebar-content.css
@@ -3,18 +3,26 @@ Theme Name: _s
 Layout: Sidebar-Content
 */
 
+.site-content {
+	direction: rtl; /* Swap the positioning of .content-area and .widget-area without affecting the source order */
+}
+
 .content-area {
-	float: right;
+	display: inline-block;
 	margin: 0 0 0 -25%;
 	width: 100%;
+	direction: ltr;
 }
 .site-main {
 	margin: 0 0 0 25%;
 }
 .site-content .widget-area {
-	float: left;
+	display: inline-block;
+	margin-left: -6px;
+	padding-left: 6px;
 	overflow: hidden;
 	width: 25%;
+	direction: ltr;
 }
 .site-footer {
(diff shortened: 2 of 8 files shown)


pixelstrolch:
Besides I had lots of annoying experiences using floats, for the big layout containers I think it's quite useful. I'm able to place secondary content left and right of my main content. The primary container stays in the first position in the HTML output. Isn't this better practice in view of responsive design and SEO?

obenland:
What's with those `6px`? Looks like a code smell.

pixelstrolch:
Sorry, I missed the direction property. Should improve my reading skills ;-) 
Obviously, I haven't use this way yet – looks interesting indeed.

iandunn:
> What's with those 6px? Looks like a code smell.

It's used to [offset the spacing the browser adds between inline elements](http://css-tricks.com/fighting-the-space-between-inline-block-elements/). I should have documented that in the stylesheet, though; I'll push an extra commit for that.

As for `px` vs `em`, it seemed like `px` would be a bit more reliable in this situation, since it's absolute. If it were a relative unit, then changing the font size could make side-by-side elements start wrapping to new lines, which would be confusing and would require the dev to dig into the box model to figure out what's going on. 

The CSS Tricks article uses `px`, and using `-6px` has worked well for me, but I've only used it in one theme so far.

On second thought, though, maybe the spacing that the browser adds is already relative. I'm guessing it's based on the value of `word-spacing`? I'll have to do some more research and testing, unless you have some thoughts on it.

iandunn:
@kovshenin pointed out the hacky-ness of [using `direction: rtl` to swap the visual order of elements](http://www.pamgriffith.net/blog/changing-element-order-with-direction-rtl), and I'm wondering if a better solution might be to use Flexbox.

That'd still be a bit hacky, since it's not really intended for structural elements, but it'd probably be better than the `direction` hack until the Grid Spec is available.

[IE doesn't support it well](http://caniuse.com/#feat=flexbox), but in this case all that would happen would be that the sidebar and content area would be swapped, which doesn't seem like a big deal to me. It could cause some annoyances for devs if other parts of the theme assumed that the sidebar was always on the ( left | right ), though. 

Thoughts?

obenland:
>> What's with those 6px? Looks like a code smell.
> It's used to offset the spacing the browser adds between inline elements.

obenland:
> > What's with those 6px? Looks like a code smell.

> It's used to offset the spacing the browser adds between inline elements.

I think we should go a different route for that. Font-size comes to mind.

> @kovshenin pointed out the hacky-ness of using direction: rtl to swap the visual order of elements, and I'm wondering if a better solution might be to use Flexbox.

Maybe there are cases where floating is not the worst solution?

iandunn:
> I think we should go a different route for that. Font-size comes to mind.

I think that has more problems than the negative-margin approach; they're [enumerated](http://css-tricks.com/fighting-the-space-between-inline-block-elements/) in the CSS Tricks article.

> Maybe there are cases where floating is not the worst solution?

It's a grey area, yeah. I personally think `direction` or Flexbox would be less hacky than floating, but it's not a big deal to me either way.

m-e-h:
Flexbox is probably the way to go for this. It just may be too soon and too new for many folks.

Here are how some other well known projects handle this.

**Suit CSS** uses font size currently but has mentioned flexbox will replace that in the near future. 
https://github.com/suitcss/components-grid/blob/master/lib/grid.css#L30

**Inuit CSS** gives the option to use negative margins but highly recommends minimizing your html instead(which is what I personally do).
https://github.com/csswizardry/csswizardry-grids/blob/master/csswizardry-grids.scss#L215

**Pure CSS** uses flexbox with negative margins and negative letter-spacing for fallbacks.
https://github.com/yahoo/pure/blob/master/src/grids/css/grids-core.css#L6

None of these are perfect but I think the Pure CSS solution is covering the most bases here though it's a little out of date. They just picked up development again so it may be worth following some of the issues they run into.

karmatosed:
For now, I am against using Flexbox. I love it, but it's too early. If people want to use it then that's fine. I'm going to for now close this up. We can review the use of Flexbox and other really cool methods as the months go by and more browser support happens.

PR #684, opened 2015-01-20
Set the default Featured Image dimensions

krishna19:


philiparthurmoore:
Hi @krishna19. Thanks a lot for this. I'm not really sure that this is super-vital to _s, though. It's just a comment and there's already supporting documentation in the DocBlock above the proposed comment. Others may be interested in having this in _s, so I'll leave this open, but I'll voice my opinion to close those out. Seems like cruft.

obenland:
I agree, let's not do that.

PR #686, opened 2015-01-23
Separating registration and enqueuing of scripts

edwardr:
I know it's not a requirement, but it does seem like a cleaner way to do this, particularly as developers will be enqueuing more scripts and styles, generally speaking.

### DIFF
--- a/functions.php
+++ b/functions.php
@@ -100,11 +100,16 @@ add_action( 'widgets_init', '_s_widgets_init' );
  * Enqueue scripts and styles.
  */
 function _s_scripts() {
+	
+	wp_register_script('_s-navigation'), get_template_directory_uri() . '/js/navigation.js', array(), '20120206', true );
+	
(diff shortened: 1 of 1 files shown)


emiluzelac:
Cleaner way is actually not to register.

Anything theme does child can override at any given time and you can add as
many scripts or styles you need without registering them first.

Not to mention that there is no need to register here 😀

Remember, less is more.

edwardr:
Ok, thanks, I will update this in my development practices.


Edward
https://www.edwardrjenkins.com

emiluzelac:
No problem 👍

obenland:
Yes, since `wp_enqueue_script()` automatically registers the script as well, you would only register a script without enqueueing it when you would want to make it available to other parts of the application, but don't necessarily need it all the time. Like most of the core-bundled scripts.

PR #689, opened 2015-01-23
Improve translation string of _s_posted_on()

onedesigns:
The output of _s_posted_on() function is split in 2 translation strings with no possibility to interchange them. Currently the output is locked to "Posted on {date} by {author}". Combining them into a single translation string allows for output like "Posted by {author} on {date}" or even "{author} wrote this on {date}" and it even improves the translation possibilities for languages with non-latin semantics.

I know that I have made some dramatic changes to established translation strings, but also adding HTML tags to translation strings isn't the most elegant solution and this pull request doesn't need to be merged as is. I propose a discussion wether prepositions like "on" and "by" actually make a semantic sense when wrapped in <span> tags together with the meta and wether the word "Posted" is directly related to the post date or makes sense even outside this context.

### DIFF
--- a/inc/template-tags.php
+++ b/inc/template-tags.php
@@ -83,16 +83,16 @@ function _s_posted_on() {
 	);
 
 	$posted_on = sprintf(
-		_x( 'Posted on %s', 'post date', '_s' ),
-		'<a href="' . esc_url( get_permalink() ) . '" rel="bookmark">' . $time_string . '</a>'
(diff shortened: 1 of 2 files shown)


fklein-lu:
I think the explanation makes sense, mind refreshing the PR please @onedesigns?

karmatosed:
This PR is super stale unfortunately. I'm for now going to close it, but please consider a refresh.

PR #691, opened 2015-01-25
Minor SASS Syntax corrections

slushman:
Found a few places where the SASS syntax could be corrected and shortened on several files. I also found some styles that were duplicates after _s moved to Normalize. I also noticed in the WP CSS Code Standards the other day that each CSS property should be on its own line, so I fixed those as well. These are all super minor - and none of these changes alter the style.css, just the SASS files.

### DIFF
--- a/sass/_normalize.scss
+++ b/sass/_normalize.scss
@@ -43,11 +43,11 @@ template {
 
 a {
 	background-color: transparent;
-}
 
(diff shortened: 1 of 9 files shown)


slushman:
Forgot to create an issue first and one change I made does change the style.css file; I didn't include that.

PR #700, opened 2015-02-14
Consistent navigation

pixelstrolch:
Handle the comment navigation the same way as the new the_post_navigation() and the_posts_navigation().


ScottSmith95:
This looks great. Good move here.

PR #703, opened 2015-02-18
Using best practices to apply the proper units to the font-size mixin. S...

taupecat:
...ee http://sass-guidelin.es/#units


emiluzelac:
That looks very nice 👍

PR #706, opened 2015-02-27
Fix child menu styling inheritance.

tkc49:


tkc49:
I forgot change "style.css"

karmatosed:
I've been going back and forth thinking about this. I think as it's up to the child theme to do this really, it probably isn't needed as a PR. For now, I'm going to close this as it's also not had any traction since March. 

Thanks for your contribution, though and we are open to considering things if you want to start and issue and look to get conversation around how to support child themes better.

PR #707, opened 2015-03-02
Removal of duplicate function

pratikshrestha:
Removed duplicate function the_posts_navigation() in inc/template-tags.php

### DIFF
--- a/inc/template-tags.php
+++ b/inc/template-tags.php
@@ -7,13 +7,13 @@
  * @package _s
  */
 
-if ( ! function_exists( '_s_posts_navigation' ) ) :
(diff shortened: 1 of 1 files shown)


philiparthurmoore:
Hi there. As you can see by this DocBlock `@todo Remove this function when WordPress 4.3 is released.` this function is needed for installations that are older than 4.1. When WordPress 4.3 is released we will remove this function from the theme itself, but until then we need to have that function in the theme so that old installs do not break. Thanks for the PR but it's not necessary just yet. Cheers.

pratikshrestha:
Thanks for your reply. I did it as lots of themes using _s that have been submitted for review in WordPress.org had two functions with same name. Most authors left it as it is :)

PR #716, opened 2015-03-24
Add bang to scss stylesheet header to prevent it being removed. Fix #715

### LINKED ISSUE #715: When sass is compiled with compressed output mode, stylesheet header is lost
Need to add a bang, as described in the [SASS docs](http://sass-lang.com/documentation/file.SASS_REFERENCE.html#comments), to the comment.

acusti:


obenland:
Opinions?

ScottSmith95:
I've had to do this, especially to prevent my build processes from removing the theme info and breaking the theme. I had concerns about this, but this seems to be a very standardised way of preventing comment removal and WordPress has no problem with it.

glueckpress:
The related issue #715 was closed as wontfix-for-now by @davidakennedy. Is it being reopened? Otherwise I don’t see the point of this PR.

acusti:
I didn’t even realize the PR was still open. Sorry about that!

PR #717, opened 2015-03-24
Menu: Renamed "Primary Menu" to "Primary Location" to avoid confusion for users

folletto:
In the tests run on WordPress.com we identified that themes are using “Menu” when they mean “Location” and this is a factor that contributes in confusing the users. 
In WordPress UI copy the “Menu” is a different object from “Location”.

For example, wp-admin tabs are labelled clearly in this way:

screen shot 2015-03-24 at 10 21 30

The result is simple. This is how it was before:

screen shot 2015-03-24 at 10 20 58

And here's after the patch:

screen shot 2015-03-24 at 10 21 52


emiluzelac:
This doesn't sound right to me. It's a menu not a location 😀

folletto:
> This doesn't sound right to me. It's a menu not a location

Uhm. Actually it's the opposite, as wp-admin (screenshot above) show. :)

emiluzelac:
Davide, WPCOM is not different from WPORG please see
https://codex.wordpress.org/Appearance_Menus_Screen#Theme_Locations

folletto:
WPcom? What about WPcom? 

This is about how the theme labels a _location_ in wp-admin (Core), which currently wrongly states "Menu" instead of "Location". And even if this wasn't a mismatch in how `wp-admin` and `nav-menus.php` (Core) works, we have tested it and confuses users.

So, what is your argument against this clear mismatch?

davidakennedy:
I'm in favor of this. It's a good, simple change that will clear up confusion for users.

samikeijonen:
I can see both wording understandable. But I have never seen theme authors use word Location and I haven't heard word Menu confusing. Do you have more info how the test run was made and by how many people?

folletto:
I was surprised as you are. :)

It was a simple live test (2 people) that was testing other flows, but that had some manipulation of Menus inside. While I agree that usually a conclusive test should have ~5 users, the fact that it emerged in both of them paired with the fact that was a trivial change made it for... a quick update. :)

As you mention, it might be equally understandable, but in the end, `wp-admin` refer to one as "Location" and the other as "Menu" quite clearly already, so even if the result was equal, it's still a tiny change that resolves a mismatch on the UI, so it's advisable. :)

obenland:
Yeah, I can see how that is confusing.

emiluzelac:
I don't know bud, renaming menus into locations seems very odd to me. I
would leave up to devs to decide.

obenland:
It's about renaming menu locations from Menu to Location.

emiluzelac:
How about just Primary without location?

samikeijonen:
Or just name the menu `Primary` like @justintadlock. No Location or Theme after word Primary. That works even better for me.

mor10:
This seems to be a question of perception versus function. "Primary Menu" indicates to the user this will be the main (primary) menu on the site. "Primary Location" explains where the menu will display in relation to the code in the theme. 

While I don't doubt the data you have, I am not sure changing the nomenclature to "Location" will make things clearer. Based on my interaction with users the term "Location" is more of a confusion. I often field questions about where a specific theme location is, and the naming convention "Primary Menu" often resolves this:

"There are two locations on the site - the primary menu area and the secondary menu area."

If anything I would urge theme developers to provide more descriptive menu names like "Header Menu" and "Footer Menu". Or eliminate the " menu" and just say "header" and "footer".

I fear "Primary Location" might be perceived by users as "somewhere on the site".

justintadlock:
I was asked about this ticket, so I'll just leave my thoughts here.

It is actually confusing when a theme author labels their nav menu locations with the word "Menu" in it.  I personally don't use "Menu" or "Location" in the label and rarely get questions from users about it.  However, I could see "Location" being added to reinforce the difference for users.

Side note: I also think it's a bit confusing for theme developers.  Believe it or not, I've had more questions from devs trying to understand the difference than users.  This comes from the function name `register_nav_menu()`.  I doubt it'll go anywhere, but it might be worth opening a core Trac ticket to get a new function name like `register_nav_menu_location()`, `register_nav_location()`, or `register_menu_location()`.

folletto:
> primary menu area and the secondary menu area

Exactly. You had to add _"area"_ to clarify. ;)

That said, to me this is an incredibly straightforward change, regardless if it's "Location", "Area" or nothing, the point is that the UI already says"Location", so I did this PR to match that. 

screen shot 2015-03-24 at 16 47 17

I think that going for just the name would work too, it's just slightly less explicit because it relies more on the page context, but would solve the same problem I outlined above. :)

And yes @justintadlock ... the confusion for developers arises from the function naming. That's a way way bigger change tho... while we can solve it right now, at least this part, quickly. ;)

mor10:
After reading my own response I realize it's actually an argument in support of this PR @folletto. I just had to spell it out for myself I guess.

karmatosed:
I'd +1 this, to me it makes far more sense for users.

glueckpress:
+1 for _Primary Location_, as it leaves less space for misinterpretation imo.

philiparthurmoore:
I'd be more in favor of _Primary_, _Secondary_, and _Footer_ or, similar. The words _Menu_ and _Location_ are quite confusing, and even become more confusing when things like Social link menus (see Twenty Fifteen) are added in.



I think the problem is less in how themes do it and more in how this "Theme Location" page is worded, but wouldn't be bothered if _Menu_ and _Location_ were ditched. Both are confusing.

philiparthurmoore:
Also, _Location_ is pretty redundant here. Here's the text on that page: "Your theme supports [number] menus. Select which menu appears in each location." No need to add _Location_ next to each menu name, IMO.

corvannoorloos:
> I'd be more in favor of Primary, Secondary, and Footer or, similar.

:+1:

fklein-lu:
I agree with Philip on this one.

folletto:
So if I count correctly:

* 5 in favour of this patch as-is
* 5 in favour of this patch, but dropping "Location" as well
* 0 against this patch

Which to me translates to: let's do this. If anyone can make a call for _with_ or _without_ Location, I can update this and we can get this merged. ;)

karmatosed:
Primary and secondary do not help with people knowing what they relate to. Whether you add location of not, they are giving weight you may not want to areas. Users also don't know what primary and secondary are. It's very much online terms that not all users at all understand.

We have a few options for what to replace:
- Call it as has been suggested what it is in case of _s : header. 
- Call it main. I feel this also is an issue as verging into online speak.

I still say my vote is for header as it's what this is. It's header navigation. This will encourage people to also then use the right naming for what it does.

philiparthurmoore:
There seems to be a few issues that this raises. I'd like to make sure that we are clear on all of them.

1. The original proposal that this pull request is submitting is to change "Primary Menu" to "Primary Location". I've given [comments above](https://github.com/Automattic/_s/pull/717#issuecomment-85955463) about why I think this does not necessarily reduce user confusion.
2. There has been an additional comment made about changing the naming of "Primary Menu" to something else, like "Header". Header might not be the best idea here, as Twenty Fifteen shows that a primary menu of a site can be in the sidebar.

_s itself has only one menu, named "Primary Menu":

```php
	// This theme uses wp_nav_menu() in one location.
	register_nav_menus( array(
		'primary' => __( 'Primary Menu', '_s' ),
	) );
```

For the specific purposes of this pull request, I do not think that we need to do anything here that does not involve dropping the additional name "Menu". "Primary Menu" or "Primary" would work well. Dropping "Location", which this current pull request does not address, would be acceptable IMO. "Location" is redundant, especially in cases where we are treating _s as a model for how menus should be declared.

Imagine a theme with four menus, a primary menu (most important), a secondary menu (less important), a social links menu (like how it's done in Twenty Fifteen) (this feels like a wonky workaround to showing social links on people's sites), and a footer menu. Which naming convention would make the most sense?

"Menu":

- Primary Menu
- Secondary Menu
- Social Menu (This one makes me uncomfortable. Social Links would be more exact.)
- Footer Menu

"Location":

- Primary Location
- Secondary Location
- Social Location (This is just weird.)
- Footer Location (Also very weird.)

Nothing:

- Primary
- Secondary
- Social
- Footer

Naming based on the specific needs of themes:

- Header
- Sidebar
- Social Links
- Footer

As you can see, each of these approaches has issues. I do not think that adding "Location" helps here, and if anything think that ditching "Location" and "Menu" are wisest for setting a good example.

philiparthurmoore:
Also, to clarify a few of these points, we need to keep in mind where this stuff is being viewed. Here is where uses see anything related to a menu:

In the Customizer:



On the old screens:

philiparthurmoore:
One final note, the more I think about this the more I do not know that we can really set a standard here. I think that words like Header, Footer, Sidebar, and Social are all simple and very easy to understand. Things like Primary, Secondary, "Menu", and "Location" feel developer-focused.

emiluzelac:
+1

folletto:
Sounds good to me, either way fixes the confusion that I mentioned above. 

Removed "Location" from this PR. Now it's just "Primary". :+1:

fklein-lu:
I'm in favor of the more plainly named locations. In the context of the Customizer, these make a lot more sense, and give the user an indication of where to look for the changes in the live preview.

karmatosed:
I still think calling this Header would be good, because it's the Header navigation. I don't really think Primary gets us any more description.

philiparthurmoore:
Somewhat related: https://core.trac.wordpress.org/ticket/33743.

karmatosed:
As this is a better solution than we have and it's sitting here, going to suggest we get a patch refresh and merge this @folletto. Are you ok sorting that patch out?

folletto:
Unfortunately I wasn't able to do it. I can resolve the conflict with the merge, but there's a problem with the way the POT file has been updated in the meantime: before it was a single line referencing a single string, while now it's getting the string from _two_ locations so they should be split first (btw, yet another reason for this string to be different, since the translations for these two could be different in different languages).

I looked around but I couldn't find any information on how to regenerate the POT file of `_s`.

This is the part I'm referring to:

```
#: functions.php:47
#: header.php:38
#@ _s
msgid "Primary Menu"
msgstr ""
```

Sorry.

davidakennedy:
@folletto I usually use Poedit to regenerate the `.pot` file. If you don't have that, maybe just submit a new PR with just the menu name change, and let us worry about the `.pot` file update.

folletto:
Ok done.

Change made, but POT not updated. 

Feel free if you prefer to edit this branch to update the POT before merging, so we get a clean merge instead of 2 PRs. ;)

sarahmonster:
This had a bit of a messy history that became problematic to cleanly rebase, so I've added a replacement PR: #872. :smiley:

Issue #230, opened 2013-05-28
Enhancement - provide translation ready .po file


idosius:
Currently there is only a readme.txt in the languages folder saying the obvious. 

Instead, the theme should come ready with a .po file that contains all the theme's strings ready for translation. It would save everyone time and allow immediate translation.

Better yet - there should be ready translations in the languages folder. I can provide a Hebrew translation if you're interested.

PR #718, opened 2015-03-26
Portuguese localization (pt_PT)

pedro-mendonca:


obenland:
Thanks for your PR, but we don't include translation files since `_s` is not a finished theme.
See #422, #230, #111, #108, #50.

pedro-mendonca:
I understand, still won't it be usefull for people to use the startup .po files in their languages also as a way to understand localization?
Thanks

PR #722, opened 2015-04-02
compress screenshot for 92% savings! bradp@067dac2

mrwweb:
I should've done this originally in 4da6d65, but better late than never!


obenland:
Actually, would you mind increasing the image size to 1200x900? The WPTRT increased the allowed size recently and we could set a good example here.

philiparthurmoore:
@obenland When was that requirement changed? Got a link? On my phone so a bit difficult to find.

obenland:
I proposed the change a few weeks back to make screenshots in the new Directory retina compatible. The handbook was only updated today: https://make.wordpress.org/themes/handbook/review/required/#screenshot

philiparthurmoore:
Awesome. Thanks for that information.

mrwweb:
Great. I had no idea. Will update shortly.

mrwweb:
Okey dokey! Bigger image and—somehow—an even smaller file size.

PR #723, opened 2015-04-02
remove redundant link :hover and :active style, shift a:focus to reset

mrwweb:
Right now, [lines 514-517 of style.css](https://github.com/Automattic/_s/blob/70b8baf2ed1a3b965aa163204df45bec8208ae1d/style.css#L514) are identical to [lines 96-99](https://github.com/Automattic/_s/blob/70b8baf2ed1a3b965aa163204df45bec8208ae1d/style.css#L96) (except for order of the two selectors).

This PR removes the second instance, deeming the style more "normalize" than "link." As part of this move, I also shifted the now-"orphaned" `a:focus` style that accompanied the second instance to the be next to the first instance of the redundant style so all normalizing link stuff is in one place.

From an accessibility perspective, I think this is better too as the remaining link styles should be edited with ease (they're just colors) but the :focus style requires a bit more care and must not just be removed.

### DIFF
--- a/sass/navigation/_links.scss
+++ b/sass/navigation/_links.scss
@@ -9,11 +9,7 @@ a {
 	&:active {
 		color: $color__link-hover;
 	}
-	&:focus {
+	&:focus { /* the :focus state is important for keyboard accessibility: http://24ways.org/2009/dont-lose-your-focus/ */
 		outline: thin dotted;
-	}
-	&:hover,
-	&:active {
-		outline: 0;
 	}
 }
--- a/style.css
+++ b/style.css
@@ -507,13 +507,8 @@ a:active {
 	color: midnightblue;
 }
 
-a:focus {
+a:focus { /* a :focus state is crucial for keyboard accessibility: http://24ways.org/2009/dont-lose-your-focus/ */
 	outline: thin dotted;
-}
-
-a:hover,
-a:active {
-	outline: 0;
 }
 
 /*--------------------------------------------------------------


obenland:
The focus styles should probably stay where they are, so normalize can easily be updated. But the hover/active styles can be removed, nice catch. Would you mind double-checking the sass files for that too?

mrwweb:
I was about to make this changed and I think I realized why this happened in the first place. If you have this

```
a:active, a:hover { outline: 0; }
a:focus { outline: 1px dotted; }
```

then an `:active` link shows the dotted outline because the selector specificity is equal so source order determines the style precedence. This is probably why someone added the second copy of the "redundant" style after the first `:focus` style.

Probably the best solution is source order (and therefore the PR actually stumbled into a good fix), but I see why you wouldn't want that for maintainability.

Another option is adding *another* style:

```
a:focus { outline: 1px dotted; }
a:focus:active { outline: 0; }
```

That works and is technically smaller and not redundant but also feels awfully "edge-casey." But at least it's clearer of what it's doing.

There is also of course this option:
```
a:active, a:hover { outline: 0 !important; }
````
But that is horrible *and* would modify the normalize styles.

I would use a second opinion. After all of this, I prefer the original PR but don't have a good suggestion for how to make it maintainable.

It looks like the same issue is in SASS. Whatever I do for the vanilla-CSS version, I'll make sure to do in SASS.

**Comment by obenland** [comment]
I'm not sure I understand. Focus styles are supposed to have an outline.

**Comment by emiluzelac** [comment]
This might also be good for accessibility no?

**Comment by emiluzelac** [comment]
The outline part to be more precise :)

**Comment by mrwweb** [comment]
The outline on `:focus` is absolutely 100% good for accessibility and should stay. In fact, I'm thinking that adding a comment to say as much in the CSS would be a good educational tool. The question is how does that interact with `:active`.

What I think people don't want is the `:focus` style to be applied when the link is also `:active`. When you are clicking a link, it gets `:focus` *and* `:active` so you'd see the outline by default. This can make the outline "flash" when clicked. However, as an accessibility feature, `:focus` is primarily for keyboard users. So I am assuming that what people want (just guessing) is best defined by this style.
```
a:focus:active { outline: 0; }
```
(I don't think the equivalent thing is an issue for `:hover`.)

Does that make more sense?

obenland:
`_s` is a starter theme, it doesn't need to make all decisions. I think removing the duplicate styles will do fine for now.

mrwweb:
I think that sounds fantastic. Thank you for being my sanity check. Update coming.

mrwweb:
Ok. New commit just removes duplicate style and adds note about a11y. Changes made both in vanilla-CSS and SASS styles.

obenland:
So it turns out those duplicate styles are there for a reason. Without them an outline is shown on completion of a mouse click, not only when the link is focused.

mrwweb:
Right. That's what I was trying to communicate above!

If we could change the source order so the :focus rule is before the normalize styles, then there's no need for the duplicate style. I'm guessing that that still isn't an option though. Even then, the styles are an incomplete "fix" as noted in the article I linked to in the most recent commit.

Would you at least consider a PR that links to an article about the important of `:focus`. Rather than the most in the most-recent PR, I'd probably link to: http://www.outlinenone.com

For background, it's worth considering [what happened when Eric Meyer originally zeroed-out `:focus` in his CSS reset](http://meyerweb.com/eric/thoughts/2011/01/03/reset-revisited/).

PR #724, opened 2015-04-04
Fix for <div id="content class="site-content">

pkinchla:
When html validating a recent site built on _s I noticed that there are two elements with duplicate classes and id's. There is a `<div>` in header.php with an id of "content" and class of "site-content". These show up again on the `<main>` element in both the single.php and page.php templates. It seemed to make sense to remove the `<div>` element. It seems to make more sense from a semantic point of view to have the id of content on the `<main>` element. I am a first time contributor to this repo so please forgive my ignorance if I am missing something.

### DIFF
--- a/footer.php
+++ b/footer.php
@@ -8,15 +8,13 @@
  */
 ?>
 
-	</div><!-- #content -->
-
-	<footer id="colophon" class="site-footer" role="contentinfo">
-		<div class="site-info">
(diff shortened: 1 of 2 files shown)


philiparthurmoore:
I'm not seeing this in [page.php](https://github.com/Automattic/_s/blob/master/page.php) or [single.php](https://github.com/Automattic/_s/blob/master/single.php).

philiparthurmoore:
By the way thanks for your first contribution. I don't think this is an issue that needs to be fixed.

pkinchla:
You are correct. I misread the class and id names. My bad.

PR #729, opened 2015-04-15
State licence of Underscores

grappler:
When generating a theme there is no mention what license Underscores is released under. The PR is based on the advice from Chip in the post "[Proper Copyright/License Attribution for Themes](https://make.wordpress.org/themes/2014/07/08/proper-copyrightlicense-attribution-for-themes/)"


davidakennedy:
This seems like a good thing to do. If we go ahead, the `style.scss` should also be updated to match.

obenland:
And it needs a trailing period. @grappler would you mind updating the PR and squashing your commits?

grappler:
It should be all good now. :smile:

grappler:
Can we merge this PR if there are no other changes needed to be made? :smile:

PR #731, opened 2015-04-17
remove double outline declaration

HeikoMamerow:
We have declaration *outline: 0* for *a:active* & *a:focus* already: 

*style.css > Line 96*
*sass > _normalize.scss > Line 48*

Once is enough ;-)


obenland:
Looks like a duplicate of #723

mrwweb:
Definitely a duplicate. What if we remove not just the duplicate `outline: 0` style but also the immediately preceding `:focus` style? I think there's a good argument that we should just follow the browsers default `:focus` style.

I also thought a [comment about the importance of `:focus` for a11y](https://github.com/Automattic/_s/pull/723#issuecomment-90269222) would be worthwhile.

PR #733, opened 2015-04-22
added menu navigation with arrow keys

tellthemachines:
Added a bit of code to navigation.js to enable keyboard navigation with the arrow keys on menus (not just the main navigation menu, but also any extra menus there may be - widget menus, etc.). This is useful for large menus with sub-menus as it enables navigation of top items without having to go through all sub-items.

### DIFF
--- a/js/navigation.js
+++ b/js/navigation.js
@@ -78,4 +78,60 @@
 			self = self.parentElement;
 		}
 	}
+
+	// menu navigation with arrow keys
+
+	var mainMenus = document.getElementsByClassName('menu');
+	var subMenus = document.getElementsByClassName('sub-menu');
+
+	var combinedMenus = [].concat(Array.prototype.slice.call(mainMenus), Array.prototype.slice.call(subMenus));
+	var cMenusLen = combinedMenus.length;
+
+	for(var m = 0; m < cMenusLen; m++) {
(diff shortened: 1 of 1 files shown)


obenland:
It could also lead to unforeseen behavior when other parts of the theme or a plugin rely on keyboard navigation. It could make sense in a finished theme, but I'm not sure it belongs in a starter theme that aims to provide the smallest common denominator.

davidakennedy:
Hey @tellthemachines!

Thanks for this pull request! It's a great idea from an accessibility standpoint, but I think it's a bit much considering we already have basic keyboard navigation in the dropdown menu and as @obenland mentioned, could create complications that we can't foresee because we're never sure how people will use _s.

But your code can still help a lot of people. The WordPress Accessibility Team is putting together a [accessible theme pattern library](https://make.wordpress.org/accessibility/tag/theme-pattern-library/), and this would be the perfect kind of thing to go there. Just open an issue to suggest it or submit a pull request. [Repository is here](https://github.com/wpaccessibility/a11ythemepatterns).

tellthemachines:
Thanks for the feedback guys! Will submit it to WP accessible theme pattern library.

davidakennedy:
Thanks!

PR #736, opened 2015-04-23
Option to remove margins from Admin Toolbar

mikehaberle:
Code that can be uncommented to remove margins added by the Admin Toolbar.

Useful for if a website has elements at the top of the page that use absolute positioning, as the toolbar can push things out of alignment or create white space.

Could be useful to some.

### DIFF
--- a/functions.php
+++ b/functions.php
@@ -70,7 +70,7 @@ function _s_setup() {
 	) );
 	
 	/*
-	 * Remove margin from HTML/Body added by Admin Bar
+	 * Remove margin from HTML/BODY added by Admin Bar
 	 * Useful if absolute positioning used in header area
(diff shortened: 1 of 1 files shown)


emiluzelac:
Would it be easier just checking the box from the users profile and remove the bar?

screen shot 2015-04-23 at 4 27 09 pm

mikehaberle:
That would remove the toolbar completely. This would leave it, but solve the spacing issue that can occur.

emiluzelac:
Yes I know that. Personally I strongly believe that themes should not make such changes. Themes are meant for design not really to "correct" core margins :)

mikehaberle:
Fair enough!

PR #738, opened 2015-04-27
Fixed typo in documentation

mjchamplin:
Line 25 overload > override


obenland:
It's in reference to something like function or method overloading, since the child theme's template will be used rather than the parent's. What would be your reason to change it?

mjchamplin:
My mistake -- I misunderstood the syntax on this one.

PR #739, opened 2015-04-29
#H0Dv3qS1 Add All Types Custom Fields to Public API Response

josiahsprague:
This adds add Types controlled custom fields to the public API response, instead of just the ones in a group named "public".


josiahsprague:
Oops, wrong branch.

Issue #740, opened 2015-05-03
Hfeed on single pages


dshanske:
I considered sending a pull request for it, but I'm not sure if there is a historic reason for it I'm missing.

<div id="page" class="hfeed site">

This means that any page, regardless of if it is a single page or not, is a feed. If I understand the specification correctly, it should only add hfeed if it is not a singular page. A single page is not technically a feed.

Wanted to ask about this before submitting because I searched around and cannot find discussion on the matter.

grappler:
This was partly discussed in https://github.com/Automattic/_s/issues/510

dshanske:
I read #510 and saw where the issue of Microformats in general was brought up and the discussion of the placement of hentry, but not any discussion of hfeed. The decision that appeared to be made about hentry was that it was a core issue, which I would agree on. But this is about hfeed, which does not appear in core.

It could be added to core, but the addition of hfeed in core raises the same concerns about the current misuse of hentry in core...that changing the behavior would break more sites than it would fix. 

Since _s doesn't have a backward compatibility concern in this case, it could be fixed going forward to address the improper application without affecting any sites.

grappler:
What would the code change that you make?

dshanske:
There are some options I can think of.

1. Make the addition of hfeed to the div optional based on if(!is_singular())
2. Add hfeed based on the same condition to the body class, which would make it easier to override in a child of a theme based on _s in the future.
3. Remove the hfeed entirely, making it the responsibility of anyone concerned enough to add it to their theme.

sixhours:
> Add hfeed based on the same condition to the body class, which would make it easier to override in a child of a theme based on _s in the future.

I like this idea most, since we already have a function to add classes to the body tag.

sixhours:
See #820 for a potential patch, if we decide on a direction.

davidakennedy:
Looks like we merged this PR, so closing this out.

PR #743, opened 2015-05-09
Apply Hfeed to Singular Posts Only

dshanske:
Addresses #740 by applying hfeed to the body only on archive and index pages.

### DIFF
--- a/header.php
+++ b/header.php
@@ -18,7 +18,7 @@
 </head>
 
 <body <?php body_class(); ?>>
-<div id="page" class="hfeed site">
+<div id="page" class="site">
 	<a class="skip-link screen-reader-text" href="#content"><?php esc_html_e( 'Skip to content', '_s' ); ?></a>
 
 	<header id="masthead" class="site-header" role="banner">
--- a/inc/extras.php
+++ b/inc/extras.php
@@ -18,6 +18,10 @@ function _s_body_classes( $classes ) {
 	if ( is_multi_author() ) {
 		$classes[] = 'group-blog';
 	}
+	// Adds hfeed to the body only on pages that are not singular
+	if (! is_singular() ) {
+		$classes[] = 'hfeed';
+	}
 
 	return $classes;
 }


grappler:
@dshanske You should use the WordPress coding standards. https://make.wordpress.org/core/handbook/coding-standards/php/

dshanske:
Assuming it was the tabbing. It has been said I am tab challenged. Expandtabs was set on vi. It looked the same from my end till I saw it on github.

grappler:
The tabs and also the spacing in `if (!is_singular() ) {`. I think you also missed a tab before the comment. Once you are able to make the changes you should squash the commits together. https://github.com/ginatrapani/todo.txt-android/wiki/Squash-All-Commits-Related-to-a-Single-Issue-into-a-Single-Commit

dshanske:
Think I fixed everything.

obenland:
We probably need to be more restrictive. `hfeed` should only be added if it's a list of posts.

dshanske:
hfeed is supposed to be a list of hentrys, hentrys being datestamped content. WordPress unfortunately, puts hentry on everything...which is another pet-peeve of mine.

But are you suggesting is_single, or something broader?

obenland:
> But are you suggesting is_single, or something broader?

I'm suggesting something more limited actually, that would only apply to screens with lists of date-stamped posts.

dshanske:
What's the rationale for that? A use case for hfeed is that a page of hentrys identified with an hfeed is a substitute for atom or rss that uses the existing html page. What you describe is the equivalent of an excerpt feed.

dshanske:
The official definition of hfeed is a collection of hentrys. Therefore, it would apply on any page that had multiple hentrys.

obenland:
> Therefore, it would apply on any page that had multiple hentrys.
Agreed.

dshanske:
Any movement on this?

davidakennedy:
We addressed here: https://github.com/Automattic/_s/pull/820 Looks like we didn't realize we already an earlier PR. Closing this out.

PR #744, opened 2015-05-10
Add a readme.txt file

jamigibbs:
This is based on the recommendation of the theme review team.

https://make.wordpress.org/themes/2015/04/29/a-revised-readme/

The goal is for a uniformly structured readme file adopted by all theme authors. Having the template ready in _s would speed up the adoption.

### DIFF
--- a/readme.txt
+++ b/readme.txt
@@ -1,0 +1,27 @@
+=== _s ===
+Contributors: automattic
+Tags: featured-images, translation-ready, custom-background, theme-options
+Requires at least: 4.0
+Tested up to: 4.2
+Stable tag: 1.0.0
+License: GPLv2 or later
+License URI: http://www.gnu.org/licenses/gpl-2.0.html
+
+Short description. No more than 150 chars.
+
+== Description ==
+Full theme description.
+
+== Frequently Asked Questions ==
+
+= A question that someone might have =
+
+An answer to that question.
+
+== Changelog ==
+
+= 1.0 =
+* Initial release
+
+== Resources ==
+ * example-script.js © 2015 Jane Doe, CC0
--- a/readme.txt
+++ b/readme.txt
@@ -1,27 +1,30 @@
 === _s ===
 Contributors: automattic
-Tags: featured-images, translation-ready, custom-background, theme-options
+Tags: translation-ready, custom-background, theme-options, custom-menu, post-formats, threaded-comments
 Requires at least: 4.0
 Tested up to: 4.2
 Stable tag: 1.0.0
 License: GPLv2 or later
 License URI: http://www.gnu.org/licenses/gpl-2.0.html
 
-Short description. No more than 150 chars.
+A starter theme called _s, or underscores.
 
 == Description ==
-Full theme description.
+
+Hi. I'm a starter theme called _s, or underscores, if you like. I'm a theme meant for hacking so don't use me as a Parent Theme. Instead try turning me into the next, most awesome, WordPress theme out there. That's what I'm here for.
 
 == Frequently Asked Questions ==
 
-= A question that someone might have =
+= Does this theme support any plugins? =
 
-An answer to that question.
+_s includes support for Infinite Scroll in Jetpack.
 
 == Changelog ==
 
 = 1.0 =
 * Initial release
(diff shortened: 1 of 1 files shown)


karmatosed:
I think having a readme.txt file is a good idea, but I don't think having empty of filler content is. For example:

"+Short description. No more than 150 chars.
+
+== Description ==
+Full theme description.
+
+== Frequently Asked Questions ==
+
+= A question that someone might have =
+
+An answer to that question.
+
+== Changelog ==
+
+= 1.0 =
+* Initial release
+
+== Resources ==
+ * example-script.js © 2015 Jane Doe, CC0 
"

I would suggest unless it's real content everything should be removed. There is no requirement or recommendation to have the entire readme.txt. Lets only include the relevant information.

jamigibbs:
I think that's a good point and it would make sense to add _s related content.

My concern though is keeping the recommended structure available for authors even when a section doesn't have any content. For example, Resources. Is there anything we could use as "filler" for the resources section? It didn't make sense to leave that blank nor did it make sense to remove it entirely (otherwise authors may not realize it should be a part of the file). So I thought perhaps it would make sense to have the entire thing a template.

grappler:
Here is a suggestion.

```
=== _s ===
Contributors: automattic
Tags: translation-ready, custom-background, theme-options, custom-menu, post-formats, threaded-comments
Requires at least: 4.0
Tested up to: 4.2
Stable tag: 1.0.0
License: GPLv2 or later
License URI: http://www.gnu.org/licenses/gpl-2.0.html

A starter theme called <code>_s</code>, or <em>underscores</em>

== Description ==
 Hi. I'm a starter theme called <code>_s</code>, or <em>underscores</em>, if you like. I'm a theme meant for hacking so don't use me as a <em>Parent Theme</em>. Instead try turning me into the next, most awesome, WordPress theme out there. That's what I'm here for.

== Frequently Asked Questions ==
= Does this theme support any plugins? =
_s includes support for Infinite Scroll in Jetpack.

== Changelog ==
= 1.0 =
* Initial release

== Resources ==
_s is based on Underscores http://underscores.me/, (C) 2012-2015 Automattic, Inc.
Underscores is distributed under the terms of the GNU GPL v2 or later.
```

jamigibbs:
+1 @grappler 

Resources would probably be listed in this way though:

```
== Resources ==
Underscores http://underscores.me/, (C) 2012-2015 Automattic, Inc.
```

grappler:
I took the license example from https://make.wordpress.org/themes/2014/07/08/proper-copyrightlicense-attribution-for-themes/

I think it would be important to mention if the code is copied or built upon and what license it uses. We could also add normalize.css
```
== Resources ==
Based on Underscores http://underscores.me/, (C) 2012-2015 Automattic, Inc., MIT
normalize.css http://necolas.github.io/normalize.css/, (C) 2012-2015 Nicolas Gallagher and Jonathan Neal, MIT
```

emiluzelac:
Should the Resources be Credits instead?

jamigibbs:
I was just considering that too @emiluzelac, looking at the example @karmatosed gave on the make post:

https://themes.trac.wordpress.org/browser/resonar/1.0.3/readme.txt
https://make.wordpress.org/themes/2015/04/29/a-revised-readme/#comment-41328

emiluzelac:
I am thinking that the original wording was an error that we carried this over :) Tammie's example link is correct and let's stick with that if possible :)

jocastaneda:
> I am thinking that the original wording was an error and we carried that over 

I'm guessing so. Keep in mind the proposal was just that. A proposal. :) The main reason for the discussion was to determine what tags are, and can be, good use for themes. The ones that seem more reasonable are `Requires at least`, `Contributors`, `Tags`, and `Tested up to`. The rest of the formatting can be up for debate. :)

@grappler's is closest to what would be ideal if it gets adopted by all. The `Resources` could be what you want it to be. `Credits`, `Licenses`, `Includes`.

karmatosed:
An outside discussion to this ticket, but requires wasn't something we were going for :) I think we have to appreciate that just having is good and what the contents are can only be recommended.

Issue #737, opened 2015-04-24
Better Coding Standards Sweep


philiparthurmoore:
After https://github.com/Automattic/_s/commit/ea8c6f41a84678d833814c392d0e66835a90576f, what follows are all the coding standards errors being triggered by _s. Some require exclusions (whitespace checks) while some require action on our part.

Actual build results can be seen here: https://travis-ci.org/Automattic/_s/jobs/59816437

```
FILE: /tmp/wordpress/src/wp-content/themes/_s/single.php

----------------------------------------------------------------------

FOUND 1 ERROR AFFECTING 1 LINE

----------------------------------------------------------------------

 21 | ERROR | [x] Line indented incorrectly; expected 3 tabs, found 4

    |       |     (Generic.WhiteSpace.ScopeIndent.IncorrectExact)

----------------------------------------------------------------------

PHPCBF CAN FIX THE 1 MARKED SNIFF VIOLATIONS AUTOMATICALLY

----------------------------------------------------------------------

FILE: /tmp/wordpress/src/wp-content/themes/_s/search.php

----------------------------------------------------------------------

FOUND 1 ERROR AFFECTING 1 LINE

----------------------------------------------------------------------

 16 | ERROR | Expected a sanitizing function (see Codex for 'Data

    |       | Validation'), but instead saw '__'

    |       | (WordPress.XSS.EscapeOutput.OutputNotSanitized)

----------------------------------------------------------------------

FILE: /tmp/wordpress/src/wp-content/themes/_s/page.php

----------------------------------------------------------------------

FOUND 1 ERROR AFFECTING 1 LINE

----------------------------------------------------------------------

 24 | ERROR | [x] Line indented incorrectly; expected 4 tabs, found 5

    |       |     (Generic.WhiteSpace.ScopeIndent.IncorrectExact)

----------------------------------------------------------------------

PHPCBF CAN FIX THE 1 MARKED SNIFF VIOLATIONS AUTOMATICALLY

----------------------------------------------------------------------

FILE: /tmp/wordpress/src/wp-content/themes/_s/inc/template-tags.php

----------------------------------------------------------------------

FOUND 49 ERRORS AFFECTING 40 LINES

----------------------------------------------------------------------

  11 | ERROR | [x] Line indented incorrectly; expected at least 1 tabs,

     |       |     found 0 (Generic.WhiteSpace.ScopeIndent.Incorrect)

  16 | ERROR | [x] Line indented incorrectly; expected 1 tabs, found

[...]

PR #745, opened 2015-05-11
Fix remaining PHPCS errors

fklein-lu:
* Add ignoring in template-tags.php and comments.php.
* Escape header text color in custom-header.php.

See #737.


philiparthurmoore:
Just a note to anyone who comes across this PR that the comment `WPCS: XSS OK` is totally legitimate and the right way of getting around false errors. See https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/issues/369 for an example.

PR #746, opened 2015-05-13
Replace _ with - in .current-page-ancestor class

fostertime:
The correct class name for .current_page_ancestor is .current-page-ancestor


fostertime:
#706 covered this issue, but is not yet merged.

PR #747, opened 2015-05-13
Remove :hover and :active states on screen-reader-text

sixhours:
We should consider removing all CSS rules that provide `:hover` and `:active` pseudo element behavior for `screen-reader-text` elements.

In IE 10/11, the `screen-reader-text` element, normally hidden, pops up when you hover your mouse over it, obscuring content — the Archives Widget drop-down menu is a good example.

According to the latest from the core a11y team: ​https://make.wordpress.org/accessibility/2015/02/09/hiding-text-for-screen-readers-with-wordpress-core/ -- we only need the `:focus` pseudo element class.

Props to @lancewillett for figuring this out.

### DIFF
--- a/style.css
+++ b/style.css
@@ -642,8 +642,6 @@ a:active {
 	overflow: hidden;
 }
 
-.screen-reader-text:hover,
-.screen-reader-text:active,
 .screen-reader-text:focus {
 	background-color: #f1f1f1;
 	border-radius: 3px;


samikeijonen:
I have always wondered is `:hover` or `:active` needed. But I have never had any problems with mouse hover on IE 10/11, meaning `screen-reader-text` doesn't pop up for me. cc: @joedolson @davidakennedy.

joedolson:
I have never used the :hover or :active value for a screen reader text block that needs to become visible, and really don't see any benefit to it. In principle, if you're going to bring something into view via :hover, it would need to already be at least *partially* visible already, which would suggest a different solution, to me. 

I'd consider removing it to be perfectly reasonable. I've never encountered a problem from it, but it's not something I can see any benefits too, either.

samikeijonen:
Let them go :+1:

lancewillett:
:+1:

PR #748, opened 2015-05-14
Added optional bootstrap and font awesome includes

carl-alberto:
A lot of us know that bootstrap is bad for wordpress based from this article
http://themeshaper.com/2014/08/19/why-bootstrap-is-a-bad-fit-for-wordpress-themes/

But there are still a lot of front end developers that loves to use underscores and bootstrap because it gives us the ability to write shorter code in implementing responsive sites by familiarizing the bootstrap grid system and has greatly increased our responsive implementation by more than two folds.

Hopefully we can include this as an optional enqueue script in benefit for all of those that loves underscores and bootstrap folks out there.

Thanks for sharing us underscores! more power to all of you guys!

### DIFF
--- a/functions.php
+++ b/functions.php
@@ -111,6 +111,14 @@ function _s_scripts() {
 
 	wp_enqueue_script( '_s-skip-link-focus-fix', get_template_directory_uri() . '/js/skip-link-focus-fix.js', array(), '20130115', true );
 
+	//uncomment the 2 lines below to use bootstrap minified via cdn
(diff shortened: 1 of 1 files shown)


emiluzelac:
There is no true benefit from this, anything beyond what `_s` already does is themers decision :)

philiparthurmoore:
This could possibly find its way as an optional part of the download on underscores.me but this doesn't belong in the core _s package.

grappler:
There is a fork of `_s` that has Bootstrap integration. https://github.com/Themekraft/_tk/

PR #750, opened 2015-05-17
Minor GPL wording update

emiluzelac:
Minor GPL wording update to reflect `GPLv2 or later` what readme already does at the beginning :)


emiluzelac:
Oh, that was super quick! Thanks :)

philiparthurmoore:
Relatively minor change. No need to stall on it. Thanks for the PR.

PR #754, opened 2015-05-19
Changed classes to page-header and page-content in content-page.php file

### LINKED ISSUE #753: Class names on content-page.php
Since the content-page.php file displays the content for a page, shouldn't the header class be page-header and the content class be page-content? Each of the other page files (archive.php, search.php, 404.php, content-none.php) all use those classes, it seems to me that content-page.php should too.

slushman:
Closes #753. Change classes for page content for consistency across all pages.


slushman:
Closed. This isn't a mistake, its how it was intended to be.

PR #755, opened 2015-05-20
Fix #543; skip-link-focus-fix.js causes unexpected outline on elements in Chrome

### LINKED ISSUE #543: Bug: skip-link-focus-fix.js causes unexpected outline on elements in Chrome

davidyezsetz:


davidyezsetz:
It looks like Travis failed because of the php-codesniffer, however this PR only contains js. Did I miss something?

PR #758, opened 2015-05-21
Moved js and sass to assets folder

LoveAndHappiness:
I believe "_s" needs an "assets" folder to make integration of 3rd partie frameworks like bootstrap and foundation easier, so that we don't need to move directories around anymore and get "_s" right out of the box.

Would love to hear your thoughts on this.

P.S. 
This is my first open source contribution, please inform me if there is a better way to do this :)


fklein-lu:
I don't think this is the right fit for _s.

LoveAndHappiness:
Why do you think it is not the right fit to move all resources into an assets folder? Doesn't the root directory get cluttered in your projects with sass, js and fonts? Other frameworks like bootstrap and laravel (elixir) have made a similar transition. So, why do you think that way?

jamigibbs:
Generally, I'm not including the Sass folder in the final product so there isn't a need for me to consolidate it into an assets or includes folder. I think file structure is mostly the developers preference.

fklein-lu:
If your project has all those needs, than feel free to choose whatever structure fits well. But I don't see the benefits of adding an extra level to a very simple starter theme.

The fact that Bootstrap or Laravel handle it this way is not really an argument for doing it.

philiparthurmoore:
I'm not convinced that this will make things better. It very much looks like a developer preference and one that doesn't make making themes any easier with _s. Thanks a ton for the discussion and pull request.

PR #760, opened 2015-05-25
improved categorized blog function

gregrickaby:
Be more explicit with inline comments, use yoda conditional, return early rather than if/else

### DIFF
--- a/inc/template-tags.php
+++ b/inc/template-tags.php
@@ -222,34 +222,33 @@ function the_archive_description( $before = '', $after = '' ) {
 endif;
 
 /**
- * Returns true if a blog has more than 1 category.
+ * Returns true if a blog has more than 1 category with posts.
  *
  * @return bool
  */
 function _s_categorized_blog() {
-	if ( false === ( $all_the_cool_cats = get_transient( '_s_categories' ) ) ) {
-		// Create an array of all the categories that are attached to posts.
-		$all_the_cool_cats = get_categories( array(
-			'fields'     => 'ids',
-			'hide_empty' => 1,
 
-			// We only need to know if there is more than one category.
-			'number'     => 2,
+	if ( false === ( $all_the_cool_cats = get_transient( '_s_hippie_categories' ) ) ) {
+
+		// Create an array of categories
+		$all_the_cool_cats = get_categories( array(
+			'hide_empty' => 1, // Only count categories with posts
+			'number'     => 2, // Only return two categories
 		) );
 
-		// Count the number of categories that are attached to the posts.
+		// Count the number of categories that are attached to the posts
 		$all_the_cool_cats = count( $all_the_cool_cats );
 
-		set_transient( '_s_categories', $all_the_cool_cats );
+		// Place number of categories into a transient
+		set_transient( '_s_hippie_categories', $all_the_cool_cats );
 	}
 
-	if ( $all_the_cool_cats > 1 ) {
-		// This blog has more than 1 category so _s_categorized_blog should return true.
-		return true;
-	} else {
-		// This blog has only 1 category so _s_categorized_blog should return false.
+	// There is only 1 category, so return false
+	if ( 1 === $all_the_cool_cats ) {
 		return false;
 	}
+
+	return true;
 }
 
 /**
--- a/inc/template-tags.php
+++ b/inc/template-tags.php
@@ -228,7 +228,7 @@ endif;
  */
 function _s_categorized_blog() {
 
-	if ( false === ( $all_the_cool_cats = get_transient( '_s_hippie_categories' ) ) ) {
+	if ( false === ( $all_the_cool_cats = get_transient( '_s_categories' ) ) ) {
 
 		// Create an array of categories
 		$all_the_cool_cats = get_categories( array(
@@ -240,7 +240,7 @@ function _s_categorized_blog() {
 		$all_the_cool_cats = count( $all_the_cool_cats );
 
 		// Place number of categories into a transient
-		set_transient( '_s_hippie_categories', $all_the_cool_cats );
+		set_transient( '_s_categories', $all_the_cool_cats );
 	}
 
 	// There is only 1 category, so return false
--- a/inc/template-tags.php
+++ b/inc/template-tags.php
@@ -232,8 +232,9 @@ function _s_categorized_blog() {
 
 		// Create an array of categories
 		$all_the_cool_cats = get_categories( array(
-			'hide_empty' => 1, // Only count categories with posts
-			'number'     => 2, // Only return two categories
+			'fields'     => 'ids', // Only query category IDs
+			'hide_empty' => 1,     // Only count categories with posts
+			'number'     => 2,     // Only return two categories
 		) );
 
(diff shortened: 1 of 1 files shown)


emiluzelac:
Was `_s_hippie_categories` done purposely? Instead of  `_s_categories` :)

gregrickaby:
Lol. Oops. :)

emiluzelac:
Sorry, Déformation professionnelle :) That happens when you do this all day every day!

gregrickaby:
If anything, it fit with the whole "All the cool cats" lingo ;)

obenland:
Any reason you removed the `fields` query argument?

gregrickaby:
Because it's neither listed nor defined here: http://codex.wordpress.org/Function_Reference/get_categories

philiparthurmoore:
Pretty sure `fields` was added in for a specific performance reason by either @kovshenin or @mjangda but I'm drawing a blank and `blame` isn't working well. Am I imagining things? Also, Codex isn't a great place for looking into functions. Core's best for this:

```php
/**
 * Retrieve the terms in a given taxonomy or list of taxonomies.
 *
 * You can fully inject any customizations to the query before it is sent, as
 * well as control the output with a filter.
 *
 * The {@see 'get_terms'} filter will be called when the cache has the term and will
 * pass the found term along with the array of $taxonomies and array of $args.
 * This filter is also called before the array of terms is passed and will pass
 * the array of terms, along with the $taxonomies and $args.
 *
 * The {@see 'list_terms_exclusions'} filter passes the compiled exclusions along with
 * the $args.
 *
 * The {@see 'get_terms_orderby'} filter passes the `ORDER BY` clause for the query
 * along with the $args array.
 *
 * @since 2.3.0
 * @since 4.2.0 Introduced 'name' and 'childless' parameters.
 *
 * @global wpdb $wpdb WordPress database abstraction object.
 *
 * @param string|array $taxonomies Taxonomy name or list of Taxonomy names.
 * @param array|string $args {
 *     Optional. Array or string of arguments to get terms.
 *
 *     @type string       $orderby           Field(s) to order terms by. Accepts term fields ('name', 'slug',
 *                                           'term_group', 'term_id', 'id', 'description'), 'count' for term
 *                                           taxonomy count, 'include' to match the 'order' of the $include param,
 *                                           or 'none' to skip ORDER BY. Defaults to 'name'.
 *     @type string       $order             Whether to order terms in ascending or descending order.
 *                                           Accepts 'ASC' (ascending) or 'DESC' (descending).
 *                                           Default 'ASC'.
 *     @type bool|int     $hide_empty        Whether to hide terms not assigned to any posts. Accepts
 *                                           1|true or 0|false. Default 1|true.
 *     @type array|string $include           Array or comma/space-separated string of term ids to include.
 *                                           Default empty array.
 *     @type array|string $exclude           Array or comma/space-separated string of term ids to exclude.
 *                                           If $include is non-empty, $exclude is ignored.
 *                                           Default empty array.
 *     @type array|string $exclude_tree      Array or comma/space-separated string of term ids to exclude
 *                                           along with all of their descendant terms. If $include is
 *                                           non-empty, $exclude_tree is ignored. Default empty array.
 *     @type int|string   $number            Maximum number of terms to return. Accepts ''|0 (all) or any
 *                                           positive number. Default ''|0 (all).
 *     @type int          $offset            The number by which to offset the terms query. Default empty.
 *     @type string       $fields            Term fields to query for. Accepts 'all' (returns an array of
 *                                           term objects), 'ids' or 'names' (returns an array of integers
 *                                           or strings, respectively. Default 'all'.
 *     @type string|array $name              Optional. Name or array of names to return term(s) for. Default empty.
 *     @type string|array $slug              Optional. Slug or array of slugs to return term(s) for. Default empty.
 *     @type bool         $hierarchical      Whether to include terms that have non-empty descendants (even
 *                                           if $hide_empty is set to true). Default true.
 *     @type string       $search            Search criteria to match terms. Will be SQL-formatted with
 *                                           wildcards before and after. Default empty.
 *     @type string       $name__like        Retrieve terms with criteria by which a term is LIKE $name__like.
 *                                           Default empty.
 *     @type string       $description__like Retrieve terms where the description is LIKE $description__like.
 *                                           Default empty.
 *     @type bool         $pad_counts        Whether to pad the quantity of a term's children in the quantity
 *                                           of each term's "count" object variable. Default false.
 *     @type string       $get               Whether to return terms regardless of ancestry or whether the terms
 *                                           are empty. Accepts 'all' or empty (disabled). Default empty.
 *     @type int          $child_of          Term ID to retrieve child terms of. If multiple taxonomies
 *                                           are passed, $child_of is ignored. Default 0.
 *     @type int|string   $parent            Parent term ID to retrieve direct-child terms of. Default empty.
 *     @type bool         $childless         True to limit results to terms that have no children. This parameter has
 *                                           no effect on non-hierarchical taxonomies. Default false.
 *     @type string       $cache_domain      Unique cache key to be produced when this query is stored in an
 *                                           object cache. Default is 'core'.
 * }
 * @return array|WP_Error List of Term Objects and their children. Will return WP_Error, if any of $taxonomies
 *                        do not exist.
 */
function get_terms( $taxonomies, $args = '' ) {
```

gregrickaby:
Ok. No problem. Added it back in.

fklein-lu:
Can you please update your fork against the current version of _s and squash all the commits in the PR?

davidakennedy:
Hey @gregrickaby, we'd love to get this in if you can please update your fork against the current version of _s and squash all the commits in the pull request. Thanks!

gregrickaby:
@fklein-lu and @davidakennedy Trying to squash now... it's my first attempt at fast-fowarding a patch, I'm pretty sure I'm doing it wrong :cry: :smiley:

davidakennedy:
Hey @gregrickaby, thanks for the update!

Looks like the build is failing because the comments need a trailing period. Also, taking a closer look, you've removed some of the explanations for how this function works. We'd like to preserve that since `_s` is also a teaching tool.

Can you update your PR with the trailing periods and more comments to explain the function?

Thanks!

karmatosed:
Can we get an update on this please and a fresh PR?

inc/template-tags.php
```diff
@@ -222,34 +222,34 @@ function the_archive_description( $before = '', $after = '' ) {
...
 		set_transient( '_s_categories', $all_the_cool_cats );
 	}
 
-	if ( $all_the_cool_cats > 1 ) {
-		// This blog has more than 1 category so _s_categorized_blog should return true.
-		return true;
-	} else {
-		// This blog has only 1 category so _s_categorized_blog should return false.
+	// There is only 1 category, so return false
+	if ( 1 === $all_the_cool_cats ) {
```
jrfnl: The adjusted conditional ignores the potential situation where `$all_the_cool_cats` would return `0` (no categories). Please adjust.

iamtakashi:
We can remove `hide_empty` because the default value is `1`.

https://codex.wordpress.org/Function_Reference/get_categories

gregrickaby:
Sure can. Closing this out, and will re-submit.

PR #762, opened 2015-05-27
Fix Codesniffer errors.

fklein-lu:
Fix code style issues to get the build to pass.

### DIFF
--- a/functions.php
+++ b/functions.php
@@ -14,7 +14,6 @@ if ( ! function_exists( '_s_setup' ) ) :
  * as indicating support for post thumbnails.
  */
 function _s_setup() {
-
 	/*
 	 * Make theme available for translation.
 	 * Translations can be filed in the /languages/ directory.
@@ -39,7 +38,7 @@ function _s_setup() {
 	 *
 	 * @link http://codex.wordpress.org/Function_Reference/add_theme_support#Post_Thumbnails
 	 */
-	//add_theme_support( 'post-thumbnails' );
+	add_theme_support( 'post-thumbnails' );
 
 	// This theme uses wp_nav_menu() in one location.
 	register_nav_menus( array(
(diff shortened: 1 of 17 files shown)


emiluzelac:
Was `require get_template_directory() . '/inc/custom-header.php';` and `add_theme_support( 'post-thumbnails' );` optional by `_s` default?

fklein-lu:
It was commented out, but I don't like commented out code, and neither does CodeSniffer. :)

emiluzelac:
Awesome!

emiluzelac:
BTW, this is also great because we have been asking authors to remove
anything that theme doesn't use.

fklein-lu:
Yes, that makes sense.

I think that we can improve this by having a better generator on underscores.me. With maybe check boxes for features you can uncheck.

emiluzelac:
That would be great!

functions.php
```diff
@@ -39,7 +38,7 @@ function _s_setup() {
 	 *
 	 * @link http://codex.wordpress.org/Function_Reference/add_theme_support#Post_Thumbnails
 	 */
-	//add_theme_support( 'post-thumbnails' );
+	add_theme_support( 'post-thumbnails' );
```
philiparthurmoore: I think this is totally fine. I'm okay with uncommenting this.

functions.php
```diff
@@ -120,7 +127,7 @@ function _s_scripts() {
 /**
  * Implement the Custom Header feature.
  */
-//require get_template_directory() . '/inc/custom-header.php';
+require get_template_directory() . '/inc/custom-header.php';
```
philiparthurmoore: I also think that uncommenting this is totally fine as well. It always struck me as weird that _s had commented code for functionality like featured post thumbnails and the custom header. Thanks for handling this.

philiparthurmoore:
Looks like a good PR.

karmatosed:
I think just including support but not giving a default size isn't so good right for post thumbnails?

fklein-lu:
There is a default size, it's 150x150.

PR #763, opened 2015-05-29
Updated Tagline markup as per W3C guidelines.

sagarjadhav:
W3C has this to say:

>h1–h6 elements must not be used to markup subheadings, subtitles, alternative titles and taglines unless intended to be the heading for a new section or subsection.

<a href="http://www.w3.org/html/wg/drafts/html/master/sections.html#headings-and-sections">W3C</a>

The reason for this is that the HTML5 document outline always interprets a heading tag as beginning a new section of content, whether sectioning content tags are wrapped around them or not.

For a while there was a work-around to this whereby `<h2>` tags could be used for a tagline if they were grouped with the main heading inside a set of `<hgroup>` tags. However, the `<hgroup>` is being removed from the HTML5 spec and its use will cause code to fail validation checks.

So now the best way to handle taglines is just to use `<div>` or `<p>` tags and apply CSS to make them look like taglines, achieving your presentation goal without negatively affecting your document outline.

You can refer this [article](http://webdesign.tutsplus.com/articles/the-truth-about-multiple-h1-tags-in-the-html5-era--webdesign-16824).

### DIFF
--- a/header.php
+++ b/header.php
@@ -25,7 +25,7 @@
 	<header id="masthead" class="site-header" role="banner">
 		<div class="site-branding">
 			<h1 class="site-title"><a href="<?php echo esc_url( home_url( '/' ) ); ?>" rel="home"><?php bloginfo( 'name' ); ?></a></h1>
-			<h2 class="site-description"><?php bloginfo( 'description' ); ?></h2>
+			<p class="site-description"><?php bloginfo( 'description' ); ?></p>
 		</div><!-- .site-branding -->
 
 		<nav id="site-navigation" class="main-navigation" role="navigation">


sagarjadhav:
Here I have attached screenshot for more details about fbf356b

post-meta-spacing

fklein-lu:
8d7f3b9 looks good.

There are three problems with fbf356b:
- Consider opening a dedicated ticket for adding spacing. This is a separate issue, and has no relation to changing the tag line markup.
- Consider using CSS instead of adding spaces in the HTML.
- Spaces at the beginning or the end of translatable strings should be avoided, as there is no way for translators to know that there are spaces.

I'm happy to merge if the code is restricted to the scope of the ticket. Thanks for updating the PR.

sagarjadhav:
@fklein-lu Thanks for your inputs. I will create new ticket for fbf356b and make necessary changes. I have updated pull request as well. Hope this is fine now.
Thanks

philiparthurmoore:
Some may consider this overkill but in general I normally only output the site description if it actually exists, i.e.:

```php
							<?php
							$description = get_bloginfo( 'description', 'display' );
							if ( ! empty( $description ) || is_customize_preview() ) : ?>
								<p class="site-description"><?php echo esc_html( $description ); ?></p>
							<?php endif; ?>
```

Something to potentially consider for _s.

fklein-lu:
We could address this in a new PR, while incorporating the changes from #651.

philiparthurmoore:
Good call.

PR #764, opened 2015-06-06
Minor inline comment update

emiluzelac:
Just very minor, I don't think that we needed that extra `-` or comments about XSS :)


fklein-lu:
The comment about XSS is there to avoid a false positive from PHP Codesniffer. So this PR does not pass the build.

emiluzelac:
Thanks I had no idea 👍

PR #771, opened 2015-06-22
Added spacing between post meta (Categories, Tags, Comments) using CSS

sagarjadhav:
I have added spacing as per following comment https://github.com/Automattic/_s/pull/763#issuecomment-112889995

See the screenshot for more details

post-meta-spacing


jamigibbs:
Would it be better to add a space before the span in _s_entry_footer?

https://github.com/Automattic/_s/blob/00e6610198a0fbed48c13765f6b532889e7a8db6/inc/template-tags.php#L110

https://github.com/Automattic/_s/blob/00e6610198a0fbed48c13765f6b532889e7a8db6/inc/template-tags.php#L126

sagarjadhav:
@jamigibbs Please look at this comment https://github.com/Automattic/_s/pull/763#issuecomment-112889995

davidakennedy:
Hey @sagarjadhav!

Thanks for the pull request here! We're going to leave this as it is because _s is meant to be unstyled. We count on the themer using _s to make the style decisions they need to, and adding these styles would be one more thing some themers have to remove/change if they have a design that calls for something different.

Thanks again for the pull request, and your past contributions!

PR #772, opened 2015-06-24
Reduce one step to search replace theme name.

rahulprajapati-rtcamp:
Hello,
Here I have tried to reduce one step for search and replace theme name using define( 'THEME_NAME', '_s' ).

### DIFF
--- a/404.php
+++ b/404.php
@@ -12,11 +12,11 @@ get_header(); ?>
 
 			<section class="error-404 not-found">
 				<header class="page-header">
-					<h1 class="page-title"><?php esc_html_e( 'Oops! That page can&rsquo;t be found.', '_s' ); ?></h1>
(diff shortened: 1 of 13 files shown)


emiluzelac:
Hi,

You can go to http://underscores.me enter the text domain only once and
generate in seconds. Patch for this is neither recommended or needed :)

rahulprajapati-rtcamp:
Yes I know that but as https://github.com/automattic/_s repo has search and replace steps for theme rename. I just tried to reduce that steps.
Thank You . :)

obenland:
It's a neat idea, but you can't use constants as textdomains with gettext.
We already have a search/replace engine in place, so there is no need to reduce the work load. That would actually be more work. :)

emiluzelac:
Not only that, all textdomain must be hardcoded string too. Also see:
https://markjaquith.wordpress.com/2011/10/06/translating-wordpress-plugins-and-themes-dont-get-clever/

Issue #774, opened 2015-06-26
Add Post Title to edit_post_link for Improved Accessibility


davidakennedy:
One small tweak we could make to improve the accessibility of `_s` is adding the post title to the `edit_post_link` and hiding it with `screen-reader-text`. We do this in a few other spots to provide more context to link text, and the case would be the same here.

Thoughts?

Pull request incoming. :smile:

karmatosed:
Sounds a good idea to me.

Daronspence:
Since it seems we're waiting on a second opinion, I second this.

karmatosed:
@davidakennedy have you got any opnions on this?

sixhours:
Fixed in #775

PR #775, opened 2015-06-26
Add post title to `edit_post_link`

davidakennedy:
This improves the accessibility of these links, and gives more context for screen reader users. Now, they know exactly where the link goes.

Escaped output with `esc_html_x` and `wp_kses`.

Resolves: #774

### DIFF
--- a/inc/template-tags.php
+++ b/inc/template-tags.php
@@ -123,7 +123,14 @@ function _s_entry_footer() {
 		echo '</span>';
 	}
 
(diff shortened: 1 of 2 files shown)


obenland:
`esc_html_x()` should be a translator comment, not a context modifier.
I also don't think we need to escape `the_title()`.

davidakennedy:
I've updated the pull request. Thanks for the feedback!

We need to remember to generate a new `.pot` file too if this is merged.

sixhours:
+1

PR #776, opened 2015-06-28
New heading structure - redux of #651

mor10:
This is a rework of the now dead #651 

Provides proper semantic structure for titles throughout theme:

In header.php: Site title set to H1 on front page, div on other pages
In functions.php: Widget title set to H2 for semantic hierarchy
In content-search.php and content.php: Heading set to H2 as page title is "Search" or "Archive title" etc.

### DIFF
--- a/functions.php
+++ b/functions.php
@@ -102,8 +102,8 @@ function _s_widgets_init() {
 		'description'   => '',
 		'before_widget' => '<aside id="%1$s" class="widget %2$s">',
 		'after_widget'  => '</aside>',
-		'before_title'  => '<h1 class="widget-title">',
-		'after_title'   => '</h1>',
+		'before_title'  => '<h2 class="widget-title">',
+		'after_title'   => '</h2>',
 	) );
 }
 add_action( 'widgets_init', '_s_widgets_init' );
--- a/header.php
+++ b/header.php
@@ -24,7 +24,11 @@
 
 	<header id="masthead" class="site-header" role="banner">
 		<div class="site-branding">
-			<h1 class="site-title"><a href="<?php echo esc_url( home_url( '/' ) ); ?>" rel="home"><?php bloginfo( 'name' ); ?></a></h1>
+			<?php if ( is_front_page() && is_home() ) : ?>
+				<h1 class="site-title"><a href="<?php echo esc_url( home_url( '/' ) ); ?>" rel="home"><?php bloginfo( 'name' ); ?></a></h1>
+			<?php else : ?>
+				<div class="site-title"><a href="<?php echo esc_url( home_url( '/' ) ); ?>" rel="home"><?php bloginfo( 'name' ); ?></a></div>
+			<?php endif; ?>
 			<p class="site-description"><?php bloginfo( 'description' ); ?></p>
 		</div><!-- .site-branding -->
 
--- a/header.php
+++ b/header.php
@@ -27,7 +27,7 @@
 			<?php if ( is_front_page() && is_home() ) : ?>
 				<h1 class="site-title"><a href="<?php echo esc_url( home_url( '/' ) ); ?>" rel="home"><?php bloginfo( 'name' ); ?></a></h1>
 			<?php else : ?>
-				<div class="site-title"><a href="<?php echo esc_url( home_url( '/' ) ); ?>" rel="home"><?php bloginfo( 'name' ); ?></a></div>
+				<p class="site-title"><a href="<?php echo esc_url( home_url( '/' ) ); ?>" rel="home"><?php bloginfo( 'name' ); ?></a></p>
 			<?php endif; ?>
 			<p class="site-description"><?php bloginfo( 'description' ); ?></p>
 		</div><!-- .site-branding -->
--- a/index.php
+++ b/index.php
@@ -18,6 +18,12 @@ get_header(); ?>
 
 		<?php if ( have_posts() ) : ?>
 
+			<?php if ( is_home() && ! is_front_page() ) : ?>
+				<header>
+					<h1 class="page-title screen-reader-text"><?php single_post_title(); ?></h1>
+				</header>
+			<?php endif; ?>
+
 			<?php /* Start the Loop */ ?>
 			<?php while ( have_posts() ) : the_post(); ?>
 
--- a/template-parts/content-search.php
+++ b/template-parts/content-search.php
@@ -11,7 +11,7 @@
 
 <article id="post-<?php the_ID(); ?>" <?php post_class(); ?>>
 	<header class="entry-header">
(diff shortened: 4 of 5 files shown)


davidakennedy:
@mor10 Thanks for taking care of an updated pull request.

This looks good from an accessibility standpoint. The one piece I see missing is that if a page is set to front and the posts page is also set to a page. In that case, the posts page does not have a `h1` at all.

This is how we [handled it in Twenty Fifteen](https://github.com/WordPress/WordPress/blob/master/wp-content/themes/twentyfifteen/index.php#L24-L28). I'd recommend that approach so that we don't have a page without a top-level heading.

mor10:
I'm not sure how you'd go about doing that from Admin, but I'll get that sorted pronto.

davidakennedy:
Thanks for the quick turnaround on the update!

This looks good to me, and I'd vote for a merge. I know that this change means we're not being the most future-friendly we can be, but I think it's smart to make for these reasons:

* [No browser or assistive technology currently supports the HTML5 document outline](http://www.paciellogroup.com/blog/2013/10/html5-document-outline/).
* [Most screen reader user navigate the web by using headings](http://webaim.org/projects/screenreadersurvey5/#finding).
* This change makes it easier to create [accessibility-ready themes](https://make.wordpress.org/themes/handbook/review/accessibility/required/) from _s. It's the last bug change we can make in a starter theme like this to help theme creators.

sixhours:
I'm in favor of this, since it helps make _s accessible-friendly from the start.

mor10:
The only notable downside to this approach is that theme devs have to explicitly specify font size, height, padding, margin, and weight for site-title and entry-title because Normalize.css doesn't reset the default headings. However, that's something that should be done by the theme dev anyway so it's not really a problem, more of a note that when this gets baked in some devs are going to be annoyed until they figure out how to do it right.

obenland:
I should have asked for a walk through at #wceu contributor day to experience myself how bad it really is. 

If no one uses the HTML5 document outline, browsers and AT don't have an incentive to add support for it. How confusing is it to people really if a page has more than one H1 heading. Why is it not confusing if there is more than one H2 heading?

karmatosed:
I'm not against this, but would like to see us not use div for header.php. We can use other more semantic markup.

mor10:
@obenland This is a weighing between accessibility and future technology. IMO it is more important that we lead on accessibility - something that is properly supported across all browsers but not in most WordPress themes - than pushing for a relatively minor HTML5 feature that the browser manufacturers and the community at large seem more or less uninterested in. 

Normally I'm all for paving cowpaths and pushing future-forward code through WordPress (RICG is one example), but in this case Accessibility trumps everything. We need proper heading structures to make it easier for people who use accessibility tools to access WordPress sites. It's been a request from the accessibility community for a long time and this PR answers that need. 

Once browser manufacturers start caring about proper HTML5 content mapping I'm all for scrapping this and pushing the standard, but for now I'll side with accessibility.

To your question about headings: The H1 heading indicates what the entire page is, and it should therefore be singular. H2 headings are sub-headings and users expect to scroll through them to find what they are looking for. For this reason it makes sense to have for example an archive page have the h1 heading "Archive for whatever" and each of the posts filed under "whatever" have their headings set to h2 for easy navigation.

mor10:
@karmatosed What do you suggest? I'm open to ideas and happy to change it, but don't see any obvious candidates. Paragraph is wrong, and none of the heading levels will help us meet the goal. The reason I chose a div is that the site-title in these circumstances is more or less irrelevant: What matters is the content title.

karmatosed:
<p> is better than a div.

mor10:
I'm going to ask one of my web standards expert friends what he would do for some outside input. Stay tuned.

karmatosed:
Why not use p like I suggest? I am aware of semantic markup and web standards.

mor10:
I'm not questioning your insight @karmatosed. I'm in disagreement on the use of a paragraph. To my understanding of web standards and semantics it is incorrect. I would just like to branch out to people outside the WordPress project to get some perspective. There may well be a standardized way of doing this we are not aware of and I'd like to find out before doing anything.

mor10:
For reference, a quick review of a pile of websites including Mic, CNN, Gov.uk, BBC, Mashable, Buzzfeed etc. all mark up the site title with `H1` on the front page and a `div` on single posts and pages. The Guardian uses `div` throughout. Gov.uk in particular is interesting because they have made it a point to be as standards compliant and accessibility compliant as possible.

As a counterpoint, A List Apart uses an H1 for the title throughout (though I'm pretty sure they do that because they are paving the cowpaths even if nobody is looking).

iamtakashi:
+1. I am totally favour of this. One thing less we have to do every time to make themes accessibility-ready.

samikeijonen:
+1 from here too.

karmatosed:
The fact is outside there isn't a standard. If you were to look at a standard then more would use p than div. Div is an artefact of older semantic markup if anything. P makes sense for styling. If that happens I'd be totally in favour of this passing as a patch.

mor10:
The problem is this is not a paragraph - it's a block of text that in the specific context we are discussing is only there for reference purposes. The markup should never be chosen because of presentational value or convenience - it has a function irrespective of how it is being presented. As for "older semantic markup", most of these sites are built by the leading minds in the industry (eg The Guardian), and most have been updated within the last 6 months. Again, this is why I'm reaching out to the wider community for input.

ronilaukkarinen:
With `.widget-title`, I agree. When it comes to `.site-title`, not so sure. My wife is deaf and blind so used to test with screen readers sometimes. I agree with @karmatosed, but I can accept the div here in the end.

mor10:
I've had a chat with a couple of different people who came back with "the difference between a `div` and a `p` in this respect is negligible at best" so I'm going to amend the pull request to a `p` to fast track this through the process.

RRWD:
Morten asked me for an opinion on this, so here's my 2 cents worth:
* Agree with the content-search.php and content.php changes
* Agree with the  `<p>` in the header
* Disagree with the H2 for widgets. It will create an excess of H2 headings on a page, this will make it difficult for screen reader users to have a good overview of the structure of a page. For the Genesis Framework (version 2.2, still beta) I solved this by giving the sidebars and footer an H2 heading (primary sidebar / secondary sidebar/ footer ), hidden by .screen-reader-text, and gave the widgets an H3.

davidakennedy:
I've merged this in because it moves things forward in a good way, and puts us in line with the latest default theme. Thanks everyone for your opinions.

@RRWD Thanks for your feedback regarding the `h2` for widgets. I think it would be good to open an issue for this, and continue the discussion. If we decide we need to change, we can iterate.

mtomas7:
I have a practical question. Before I was sanitizing the title via functions.php:

```$title = esc_attr($title);```

Now we esc_url the title in the content.php. Should I remove esc_attr because it is not needed?

Thank you!

Issue #780, opened 2015-06-30
Improve links to WordPress Documentation in comments at the top of files.


PJ-Finlay:
I'm starting out with WordPress and I've been using this theme as a starter theme for the website I'm making, and I've found the short descriptions of what a file does and the links to the proper documentation on codex.wordpress.org very helpful. However this is not universally present throughout the project. For example in template-parts  content-search.php and content-none.php have a link to the template hierarchy page in the WordPress documentation while the other three files don't. I would like to create a pull request for this. Would my contribution be welcome? Thanks.

obenland:
Absolutely!

PJ-Finlay:
Thanks I'll try to have a pull request somewhat quickly.

davidakennedy:
Looks like the associated pull request for this got merged, so closing it out. Thanks!

PR #781, opened 2015-07-01
Updated documentation in comments at the top of files.

PJ-Finlay:
I tried to standardize and expand the documentation and links to
codex.wordpress.org in the comments at the top of most files. I mentioned this in issue #780.

### DIFF
--- a/404.php
+++ b/404.php
@@ -2,6 +2,8 @@
 /**
  * The template for displaying 404 pages (not found).
  *
+ * @link https://codex.wordpress.org/Creating_an_Error_404_Page
+ *
  * @package _s
  */
 
--- a/archive.php
+++ b/archive.php
@@ -2,7 +2,7 @@
 /**
  * The template for displaying archive pages.
  *
- * Learn more: http://codex.wordpress.org/Template_Hierarchy
+ * @link https://codex.wordpress.org/Template_Hierarchy
  *
  * @package _s
  */
--- a/comments.php
+++ b/comments.php
@@ -2,8 +2,10 @@
 /**
  * The template for displaying comments.
  *
- * The area of the page that contains both current comments
+ * This is the template that displays the area of the page that contains both the current comments
  * and the comment form.
+ *
+ * @link https://codex.wordpress.org/Template_Hierarchy
  *
  * @package _s
  */
--- a/footer.php
+++ b/footer.php
@@ -2,7 +2,9 @@
 /**
  * The template for displaying the footer.
  *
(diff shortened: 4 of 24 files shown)


emiluzelac:
We should actually go with full standard comment block instead:

https://pear.php.net/manual/en/standards.header.php 

For example we have @see while it should be @link instead.

PJ-Finlay:
I didn't think to look into this, but WordPress has a standard for file headers that we should be following. https://make.wordpress.org/core/handbook/inline-documentation-standards/php-documentation-standards/#6-file-headers

emiluzelac:
Yep, pretty much the same thing 👍

PJ-Finlay:
I'll start on that.

PJ-Finlay:
I updated the file header.

obenland:
Let's keep trailing periods in comments for now.
Should it be `@link` or `@see`?

PJ-Finlay:
Should I add a commit for adding the trailing periods.

emiluzelac:
@obenland  `@link` would be more appropriate. 

Link to additional information for the function/method.
For an external script/library, links to source.

Not to be used for related functions/methods; use `@see` instead.

PJ-Finlay:
I added the trailing periods but I left all of the links as @link

grappler:
Looks good :smile:

PJ-Finlay:
I'm new to open source and am not entirely clear on how this works. It seems that all of the issues with my pull request have been resolved, but it still has not been merged. How does this actually happen, sorry for the lack of understanding.

obenland:
No need to apologize :)
Most people working on `_s` do it on a volunteer basis, so it's possible that it can take a few days before someone who can merge the PR has time to do a final review and actually merge it.

@drewapicture: In your opinion, should we use `@link` or `@see` in this instance?

DrewAPicture:
@obenland: The uses of `@link` are correct.

Side note: All the DevHub and Codex links should be https.

PJ-Finlay:
Should I update that?

obenland:
Yes please. :)

It would also be great if you could squash your commits after you added more. [Here is a great manual](http://eli.thegreenplace.net/2014/02/19/squashing-github-pull-requests-into-a-single-commit/) that explains how to do it. Thanks!

PJ-Finlay:
I switched the links to HTTPS

PJ-Finlay:
I was wondering if my someone could merge my branch, everyone seems to be in agreement on the changes that I made. Thanks.

Issue #768, opened 2015-06-15
Site title color in the Customizer


samikeijonen:
There is probably a ticket for this but could not find it.

Should `.site-title` be `.site-title a` in `customizer.js` so that color change work in live preview? Otherwise link color in `style.css` can overwrite the site title link color in live preview.

https://github.com/Automattic/_s/blob/master/js/customizer.js#L28

carl-alberto:
I was also not able to find an open ticket about that issue so I have added a pull request for it in here: https://github.com/Automattic/_s/pull/785

davidakennedy:
Fixed here: https://github.com/Automattic/_s/pull/785

PR #785, opened 2015-07-27
Fix customizer header text color

carl-alberto:
This is fix from the issue raised in here: https://github.com/Automattic/_s/issues/768

### DIFF
--- a/js/customizer.js
+++ b/js/customizer.js
@@ -1,4 +1,6 @@
 /**
+ * customizer.js
+ *
  * Theme Customizer enhancements for a better user experience.
  *
  * Contains handlers to make Theme Customizer preview reload changes asynchronously.
@@ -27,8 +29,10 @@
 			} else {
 				$( '.site-title, .site-description' ).css( {
 					'clip': 'auto',
-					'color': to,
 					'position': 'relative'
+				} );
+				$( '.site-title a, .site-description' ).css( {
+					'color': to
 				} );
 			}
 		} );


davidakennedy:
Hey @carl-alberto!

Thanks for the pull request! This is something we definitely should fix.

Your PR has an issue though:

* When you try to hide the header text via the Display Header Text option in the Customizer, the site title doesn't come back when you hide, then unhide. There are style issues created because of conflicting styles being applied to `.site-title` and `.site-title a`. I think we'd have to go with something like this:

```
// Header text color.
	wp.customize( 'header_textcolor', function( value ) {
		value.bind( function( to ) {
			if ( 'blank' === to ) {
				$( '.site-title, .site-description' ).css( {
					'clip': 'rect(1px, 1px, 1px, 1px)',
					'position': 'absolute'
				} );
			} else {
				$( '.site-title, .site-description' ).css( {
					'clip': 'auto',
					'position': 'relative'
				} );
				$( '.site-title a, .site-description' ).css( {
					'color': to
				} );
			}
		} );
	} );
```

Do you want to refresh your PR and test more?

carl-alberto:
Hi @davidakennedy 

Thanks for giving me a change to get involved with this awesome starter theme. Also thanks for pointing that bug out.

I have tested and that "Blog Title" should now be displaying properly when the "Display Header Text "option in the Customizer is toggled on/off.

Issues another PR, hopefully it passes :)

Thanks

carl-alberto:
Hi @davidakennedy 

I was also wondering if it is a good idea to instead of replacing the code manually to make the header logo work, maybe that "Display Header Text "option in the Customizer would serve as the toggle to enable it ? If that is good, maybe we can create a separate PR for that?

Thanks!

davidakennedy:
Hey @carl-alberto Thanks for the updated PR!

I'm not a fan of also hiding the `.site-title a` because it doesn't need to be hidden. It's already in a container that is hidden. It's a small detail, but no sense adding CSS where it doesn't need to be. :smile:

> I was also wondering if it is a good idea to instead of replacing the code manually to make the header logo work, maybe that "Display Header Text "option in the Customizer would serve as the toggle to enable it ? If that is good, maybe we can create a separate PR for that?

Sorry, I don't quite understand what you mean here. Are you talking about the Custom Header?

carl-alberto:
Hi @davidakennedy 

Yes, IMHO 90% of the sites I worked with has a custom header image rather than text so I guess other developers has the same case as mine, but I'm not yet sure.

So instead of copying _s/inc/custom-header.php:7-12 to _s/header.php, would it be better if it would be an option from the customizer to add a logo and toggle it on/off?

Cheers!

davidakennedy:
> So instead of copying _s/inc/custom-header.php:7-12 to _s/header.php, would it be better if it would be an option from the customizer to add a logo and toggle it on/off?

Thanks for explaining. This is not something we would not add to _s. It's a starter theme, so we'd rather give theme authors the choice of whether or not to or how they would want to implement a custom header or logo.

Let's keep focused on fixing the color issue with the header text. Thanks for sharing your idea though!

m-e-h:
Why not just add `.site-title > a { color: inherit; }` to the stylesheet?

davidakennedy:
> Why not just add `.site-title > a { color: inherit; }`to the stylesheet?

That's not a bad idea. And something I wouldn't be opposed to. If we did that, it might be a good idea to add a comment as to why we're doing that. But I'd rather fix the issue where the issue takes place – in the `customizer.js` file.

What do others think?

carl-alberto:
sounds good to me :)

m-e-h:
Actually, what I proposed above wouldn't really make sense unless `custom-header.php` was changed to match `customizer.js`. So unless you're up to doing that, probably best to just change the js.

Personally, link styles drive me crazy. Adding inherit to a few select `a`s lets me ignore them. :hand:

davidakennedy:
@m-e-h Yeah, I doubt we'll do too much there with the Header WP Admin link now redirecting to the Customizer. So let's focus on the `customizer.js` file.

m-e-h:
I was just thinking the customizer part of the custom header https://github.com/Automattic/_s/blob/master/inc/custom-header.php#L70
If it's using `.site-title a` to color the text, the `customizer.js` should probably match or vise versa?
That's probably why this went so long unnoticed. Everything works fine once you save.

@davidakennedy The js code you posted at the beginning of this thread is all thats needed, right?
@carl-alberto just use that to revise...

carl-alberto:
Hi @davidakennedy & @m-e-h 
Thanks for the inputs, I have updated my PR reflecting @davidakennedy 's suggestion 
Tested it and it should be working even after toggling the "Display Header Text" checkbox
Please verify on your end.
Thanks!

karmatosed:
@davidakennedy - you seem on top of this, what do we need to do to get this progressed?

davidakennedy:
@carl-alberto Thanks for the update!

It looks like this PR got merged in, which dealt with a similar issue:

https://github.com/Automattic/_s/pull/677

I think it would be good to move to my original code above. This:

```
// Header text color.
    wp.customize( 'header_textcolor', function( value ) {
        value.bind( function( to ) {
            if ( 'blank' === to ) {
                $( '.site-title, .site-description' ).css( {
                    'clip': 'rect(1px, 1px, 1px, 1px)',
                    'position': 'absolute'
                } );
            } else {
                $( '.site-title, .site-description' ).css( {
                    'clip': 'auto',
                    'position': 'relative'
                } );
                $( '.site-title a, .site-description' ).css( {
                    'color': to
                } );
            }
        } );
    } );
```

Why? Because that would match how we do it in `custom-header.php`. I think it's better if we attach the `clip` to the outer most element, the `.site-title` and not the link nested in it. Does that makes sense?

Are you up for updating your PR again so I can merge? I'd love for you to get the props for this. :smile:

carl-alberto:
Cool @davidakennedy, I see that you are referring to custom-header.php 

I created a separate branch for this instead of the old branch that we are working with this issue so it would reflect the latest changes from the master branch that was merged in

https://github.com/Automattic/_s/pull/839

Thanks

carl-alberto:
@davidakennedy sorry for the multiple PRs to get this multiple commits squashed, I hope I got this PR correctly squashed https://github.com/Automattic/_s/pull/785/commits Thanks

davidakennedy:
@carl-alberto Got it! :rocket: Thanks for sticking with this, and getting it fixed. :smile: Congrats on your first contribution to _s!

carl-alberto:
@davidakennedy Awesome! Thanks! hopefully I can contribute more to _s and the WP community! :)

PR #789, opened 2015-08-05
Add style.css missing main font family

corvannoorloos:
Hello,

I believe the [main font family](https://github.com/Automattic/_s/blob/master/sass/typography/_typography.scss#L7) has been overlooked to add to the CSS style sheet when including support for Sass.

### DIFF
--- a/style.css
+++ b/style.css
@@ -272,6 +272,7 @@ input,
 select,
 textarea {
 	color: #404040;
+	font-family: sans-serif;
 	font-size: 16px;
 	font-size: 1rem;
 	line-height: 1.5;


davidakennedy:
Hey @corvannoorloos!

That variable is [located here](https://github.com/Automattic/_s/blob/master/sass/variables-site/_typography.scss#L1). Is that what you're looking for? Is there a bug you need to report?

corvannoorloos:
Hello @davidakennedy

But it also should be around here https://github.com/Automattic/_s/blob/master/style.css#L274.

I mean, not only to be added to the Sass files, but also available in the CSS style sheet?

Issue #784, opened 2015-07-25
Resolve strict comparison warnings


grantpalin:
I've been working on a theme based off of _s. Running PHP_CodeSniffer with the WordPress checks yields numerous results, including several warnings on using strict comparisons (`===` and `!==` instead of `==` and `!=`). I decided to run the check against a fresh download from here.

I ran the following command, which limits to just the relevant sniff:

    phpcs --standard=WordPress --sniffs=WordPress.PHP.StrictComparisons .

And the results:

    FILE: C:\Users\Grant\Downloads\_s-master\_s-master\comments.php
    ----------------------------------------------------------------------
    FOUND 0 ERRORS AND 1 WARNING AFFECTING 1 LINE
    ----------------------------------------------------------------------
     73 | WARNING | Found: !=. Use strict comparisons (=== or !==).
    ----------------------------------------------------------------------
    
    FILE: ...\Users\Grant\Downloads\_s-master\_s-master\inc\custom-header.php
    ----------------------------------------------------------------------
    FOUND 0 ERRORS AND 2 WARNINGS AFFECTING 2 LINES
    ----------------------------------------------------------------------
     49 | WARNING | Found: ==. Use strict comparisons (=== or !==).
     58 | WARNING | Found: ==. Use strict comparisons (=== or !==).
    ----------------------------------------------------------------------
    
    FILE: ...\Users\Grant\Downloads\_s-master\_s-master\inc\template-tags.php
    ----------------------------------------------------------------------
    FOUND 0 ERRORS AND 1 WARNING AFFECTING 1 LINE
    ----------------------------------------------------------------------
     106 | WARNING | Found: ==. Use strict comparisons (=== or !==).
    ----------------------------------------------------------------------
    
    FILE: ...\Downloads\_s-master\_s-master\template-parts\content-search.php
    ----------------------------------------------------------------------
    FOUND 0 ERRORS AND 1 WARNING AFFECTING 1 LINE
    ----------------------------------------------------------------------
     16 | WARNING | Found: ==. Use strict comparisons (=== or !==).
    ----------------------------------------------------------------------
    
    FILE: ...s\Grant\Downloads\_s-master\_s-master\template-parts\content.php
    ----------------------------------------------------------------------
    FOUND 0 ERRORS AND 1 WARNING AFFECTING 1 LINE
    ----------------------------------------------------------------------
     14 | WARNING | Found: ==. Use strict comparisons (=== or !==).
    ----------------------------------------------------------------------

[...]

Issue #632, opened 2014-10-15
does get_comments_number() need to be compared to '0'?


SeanChDavis:
So I was taking a look around Twenty Fifteen in trunk today and noticed in the single and page template files that the comments_template() was wrapped in a slightly different conditional than what I'm used to with _s. 

Excluding the `comments_open()` function, instead of `'0' != get_comments_number()`, the function simply stands alone. `get_comments_number()` always returns a numerical value according to the codex. That means when it returns 0, which is false, the function acts a boolean and the condition does not pass.

So... is the `not equal to` comparison to '0' just extra? Is there a reason why it has made it this far? School me if I'm missing something! Props to @sumobi to making sure I understood my suspicions before asking.

PR #792, opened 2015-08-07
Update comparisons to resolve the strict comparison warnings

grantpalin:
See discussion at https://github.com/Automattic/_s/issues/784

### DIFF
--- a/comments.php
+++ b/comments.php
@@ -72,7 +72,7 @@ if ( post_password_required() ) {
 
 	<?php
 		// If comments are closed and there are comments, let's leave a little note, shall we?
-		if ( ! comments_open() && '0' != get_comments_number() && post_type_supports( get_post_type(), 'comments' ) ) :
+		if ( ! comments_open() && get_comments_number() && post_type_supports( get_post_type(), 'comments' ) ) :
 	?>
 		<p class="no-comments"><?php esc_html_e( 'Comments are closed.', '_s' ); ?></p>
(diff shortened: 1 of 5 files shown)


comments.php
```diff
@@ -72,7 +72,7 @@
 
 	<?php
 		// If comments are closed and there are comments, let's leave a little note, shall we?
-		if ( ! comments_open() && '0' != get_comments_number() && post_type_supports( get_post_type(), 'comments' ) ) :
+		if ( ! comments_open() && get_comments_number() && post_type_supports( get_post_type(), 'comments' ) ) :
 	?>
```
philiparthurmoore: Why'd you take out `'0' !=` instead of adding in `'0' !==`?

obenland: I think I started doing that in default themes at some point. We don't really need a strict comparison here, we just need to know if it's truthy.

grantpalin: I did that based on the conclusion from https://github.com/Automattic/_s/issues/632

davidakennedy:
Tested these out, and everything looks good.

mtomas7:
Twenty Fifteen uses:

```if ( ! display_header_text() ) :```

instead of:

```if ( 'blank' === $header_text_color ) :```

Should we go with the 2015 method?

davidakennedy:
@mtomas7 Probably a good thing to consider. Feel free to open an issue or pull request to get more eyes on it. :smile:

PR #793, opened 2015-08-08
Fixing wp_title in extras

blacRose:
It was checking if we’re on an older version than 4.1, and gets the
full title passed to the function, and appends.

### DIFF
--- a/inc/extras.php
+++ b/inc/extras.php
@@ -23,7 +23,7 @@ function _s_body_classes( $classes ) {
 }
 add_filter( 'body_class', '_s_body_classes' );
 
-if ( version_compare( $GLOBALS['wp_version'], '4.1', '<' ) ) :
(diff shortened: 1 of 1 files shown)


obenland:
> It was checking if we’re on an older version than 4.1, and gets the full title passed to the function, and appends.

Correct, that is what the code is supposed to do. If we're on 4.1 or later we'll use `add_theme_support( 'title-tag' );` and don't care about the `wp_title` callback, because the built-in feature does that for us.

We also want to only append the blog name as to not overwrite the true title. We want to enhance it, not replace it.

philiparthurmoore:
Not a relevant PR anymore. See https://github.com/Automattic/_s/pull/798 and https://github.com/Automattic/_s/commit/8e9678b0bf749096beccbcd88311ad5c988362d8.

PR #794, opened 2015-08-11
Fixed different classes of widget titles in 404 page.

Emetrop:
There is an issue that default WP widget title class is 'widgettitle' instead of 'widget-title' which is used everywhere in _s. So I just added argument to the all 'the_widget' function calls with 'widget-title' classes.

### DIFF
--- a/404.php
+++ b/404.php
@@ -22,7 +22,14 @@ get_header(); ?>
 
 					<?php get_search_form(); ?>
 
-					<?php the_widget( 'WP_Widget_Recent_Posts' ); ?>
+					<?php the_widget(
+						'WP_Widget_Recent_Posts',
+						array(),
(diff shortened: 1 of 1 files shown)


karmatosed:
Thanks for pointing this out. I've talked with a few people and we have come to the same conclusion. As we are loading a core widget, that's why we use that format. In this sense it makes sense. You will note our selectors call both classes. If you would like to see something like this changed throughout WordPress, you are welcome to see if it's something core would be interested in.

karmatosed:
For reference, you can start here to make a ticket for core to consider: https://core.trac.wordpress.org. There is a great doc here to take a look at: https://make.wordpress.org/core/handbook/tutorials/trac/opening-a-ticket/

PR #796, opened 2015-08-16
Adding usual Schema for WebPage

g-kanoufi:
WebPage schema on header, footer, mainContent, search, archive and each template parts


obenland:
See https://github.com/Automattic/_s/search?q=schema&type=Issues.

g-kanoufi:
Sorry about that i should have investigated first, but as i added those on my last use of _s i though i would share it.

Cheers.

PR #802, opened 2015-08-19
Remove remaining caption margin-top and margin-bottom

corvannoorloos:
Previously discussed here https://github.com/Automattic/_s/commit/3d6bd4df3983e3cbd0ce5f3f6629419d6d84035a.

Besides some spacing and sort-order I believe this is the last remaining difference between the current Sass and CSS files.


corvannoorloos:
Thanks :smile:

Issue #688, opened 2015-01-23
.editorconfing


oriSomething:
i think [editorconfig](http://editorconfig.org/) can be great to this project.
to those who doesn't know what it is. it's a file that forces your favorite code editor to follow some rules for some files by your definition.
for example:
- force tabs / space as indentation
- trim trailing white space
- force 2 or 4 space indentation
- ...

emiluzelac:
This would fall into a personal preference actually :)

oriSomething:
on the other hand, right now, everything is tabs so if someone uses editor's default settings of spaces he gon'na have mixed whitespace types on the document which could lead also to a mess.
also default of trailing white space it's always good for git and i'm sure not every one who uses this project is using trailing white space by default

karmatosed:
I would absolutely say this is for personal choice not to be included in the project itself.

corvannoorloos:
I'd say it has absolutely nothing to do with personal choice, but whether or not `_s` authors like to encourage people even better to use WordPress' coding standards when developing on this specific project (note, I'm not talking about any forks).

As is currently encouraged by the Travic CI setup, and similar as is handled with WordPress, Jetpack and almost any other mayor programming project.

It remains of course to the `_s` authors whether or not they'd like to include it.

aryaprakasa:
Agreed! personal preference/choice is for production version  (generated via underscores.me).

But for the main branch of `_s`, which are open for public to pull request should be standardized. Just like WordPress [did](https://core.trac.wordpress.org/browser/trunk/.editorconfig).

We know that some editors require to install the plugin first and some editors didn't support `.editorconfig`. But this will help developers adhere `_s` style guide, which is based on WordPress Coding Standards.

PR #803, opened 2015-08-21
Add .editorconfig file

noskov:
This can help contributors to be consistent in their coding styles. For more info see [http://editorconfig.org/](http://editorconfig.org/). Thanks.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,15 @@
+# http://editorconfig.org/
+
+root = true
+
+[*]
+charset = utf-8
+end_of_line = lf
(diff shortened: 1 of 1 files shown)


iamdmitrymayorov:
I believe it was already discussed in #688.

sarahmonster:
As per #688, this is something we've already discussed and decided against. This is a personal preference and we don't want to force people to develop a particular way.

If someone wants to use an editorconfig, it's easy to find one elsewhere that follows the WordPress coding standards and to put it wherever makes most sense for their development setup.

noskov:
Ok.

PR #804, opened 2015-08-21
Add -moz-appearance as well

mkrdip:
```
-webkit-appearance: value; /* Chrome & Safari */
-moz-appearance: value; /* Firefox */
```

### DIFF
--- a/style.css
+++ b/style.css
@@ -195,6 +195,7 @@ html input[type="button"],
 input[type="reset"],
 input[type="submit"] {
 	-webkit-appearance: button;
+	-moz-appearance: button;
 	cursor: pointer;
 }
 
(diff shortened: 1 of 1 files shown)


sarahmonster:
I'm not certain about this, but this is likely because I've never found much occasion to use `-moz-appearance`, because Firefox tends to render input elements in expected ways. 

What does adding this property achieve here?

corvannoorloos:
Even though it's a valid point, my preference would be to leave [normalize.css](https://github.com/necolas/normalize.css) as is.

philiparthurmoore:
Probably best to maintain parity with normalize here.

karmatosed:
I'd say lets leave normalize as it is for now. It's very much a personal choice to use this. You are as with anything _s related welcome to fork and create your own version. That's the joy of _s.

Issue #799, opened 2015-08-19
the_archive_title exists since wp 4.1.0


widoz:
In template-tags.php there is a function the_archive_title with a todo, to remove it after the 4.3 is released. Btw, the function exists since 4.1.0, which is the differences since 4.3?

iamdmitrymayorov:
Themes are required to support two latest WordPress versions. Since this function was introduced in core in 4.1.0, it is no longer needed in template-tags.php. Correct me if I'm wrong.
Same thing with:
- the_posts_navigation()
- the_post_navigation()
- the_archive_description()

widoz:
Ah, didn't know the required support. Thanks for the info.

obenland:
@karmatosed Is it the current version and the one prior or the current version and two versions back? I thought it was the latter.

philiparthurmoore:
Looking at [this](https://developer.wordpress.org/themes/getting-started/setting-up-a-development-environment/#supporting-older-versions-of-wordpress):

> It’s standard practice for WordPress themes to support at least two versions back to ensure a minimum of backward compatibility. For example, if the current version of WordPress is at 4.2, then you should also make sure that your theme works well in versions 4.1 and 4.0 as well.

sarahmonster:
It's the current version and two versions back, which means 4.1, 4.2, and 4.3. :)

This means that these functions, introduced in 4.1, are a-okay to rely on now.

sarahmonster:
Closed in #800

PR #805, opened 2015-08-21
Remove _s_wp_title function

deadlyhifi:
No longer required under 2 versions back support requirements.


deadlyhifi:
[Support level mentioned here](https://github.com/Automattic/_s/issues/799#issuecomment-133422254)

PR #806, opened 2015-08-21
Translated to Spanish 

bobbylechuga:
Hello. Great idea. I made a translation into Spanish. Regards.


karmatosed:
We won't be adding translations into _s. There has been talk about this before, it's a starter theme and as a result translations would have to be generated anyway. You are more than welcome to offer them yourself in a fork or even to create a plugin. But, for at least the foreseeable future _s won't come with translations.

PR #809, opened 2015-08-22
Header link update and capitalization

emiluzelac:
Itty-bitty update for header comment block and new functions link.

### DIFF
--- a/functions.php
+++ b/functions.php
@@ -71,7 +71,7 @@ function _s_setup() {
 		'link',
 	) );
 
-	// Set up the WordPress core custom background feature.
+	// Setup the WordPress core custom background feature.
 	add_theme_support( 'custom-background', apply_filters( '_s_custom_background_args', array(
 		'default-color' => 'ffffff',
 		'default-image' => '',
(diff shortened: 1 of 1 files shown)


emiluzelac:
Also changed `Set up` to `Setup`, it reads much better and to reflect https://github.com/WordPress/WordPress/blob/master/wp-content/themes/twentyfifteen/functions.php#L109

obenland:
Setup is one word when it is a noun (e.g., “it was a setup!”) or an adjective (e.g., “follow the setup instructions”). It is two words—set up—when it functions as a verb (e.g., “I'm going to set up the computer”).

I also don't think we need to capitalize the word `functions` here.

Maybe we should do a documentation link sweep at some point and update them all to point to handbook pages instead of doing this one by one.

emiluzelac:
Updated. Should we also change https://github.com/WordPress/WordPress/blob/master/wp-content/themes/twentyfifteen/functions.php#L109 too?

All right, no more link updates. We can wait till everything is done, sorry!

emiluzelac:
Thanks!

PR #811, opened 2015-08-23
Adding _s_comment_nav() template tag.

iamdmitrymayorov:
This PR is inspired by Twenty Fifteen. I think we can add a template tag for comment navigation. It improves overall readability of comments.php and follows DRY approach. The main downside I can think of is the lack of ID on a *nav* tag. However this is exactly how it is done in Twenty Fifteen. If ID is there for a specific reason I can update a PR and add a parameter to template tag function. Something like this:
```
_s_comment_nav( $id = '' );
```

### DIFF
--- a/inc/template-tags.php
+++ b/inc/template-tags.php
@@ -7,6 +7,31 @@
  * @package _s
  */
 
+if ( ! function_exists( '_s_comment_nav' ) ) :
+/**
+ * Prints comment navigation.
+ */
+function _s_comment_nav() {
+	 // Are there comments to navigate through?
+	if ( get_comment_pages_count() > 1 && get_option( 'page_comments' ) ) : ?>
+		<nav class="navigation comment-navigation" role="navigation">
+			<h2 class="screen-reader-text"><?php esc_html_e( 'Comment navigation', '_s' ); ?></h2>
+			<div class="nav-links">
+				<?php
+					if ( $prev_link = get_previous_comments_link( __( 'Older Comments', '_s' ) ) ) :
+						printf( '<div class="nav-previous">%s</div>', $prev_link );
+					endif;
+
+					if ( $next_link = get_next_comments_link( __( 'Newer Comments', '_s' ) ) ) :
+						printf( '<div class="nav-next">%s</div>', $next_link );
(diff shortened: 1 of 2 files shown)


iamdmitrymayorov:
Updating the PR according to the failed check.

iamdmitrymayorov:
Looks good now.

obenland:
We should probably wait for https://core.trac.wordpress.org/ticket/30589

iamdmitrymayorov:
Yes, now I think we should wait too. Makes sense.

inc/template-tags.php
```diff
@@ -7,6 +7,31 @@
...
+ */
+function _s_comment_nav() {
+	 // Are there comments to navigate through?
+	if ( get_comment_pages_count() > 1 && get_option( 'page_comments' ) ) : ?>
+		<nav class="navigation comment-navigation" role="navigation">
+			<h2 class="screen-reader-text"><?php esc_html_e( 'Comment navigation', '_s' ); ?></h2>
+			<div class="nav-links">
+				<?php
+					if ( $prev_link = get_previous_comments_link( __( 'Older Comments', '_s' ) ) ) :
+						printf( '<div class="nav-previous">%s</div>', esc_html( $prev_link ) );
```
jrfnl: If this would in some form or other be merged at some point, mind that the following will need changing here and below:
`__( 'Older Comments', '_s' )` => `esc_html__( 'Older Comments', '_s' )`
`esc_html( $prev_link )` => `$prev_link`

(as other wise the `<a href=...` would become `&lt;a href=...` etc)

iamdmitrymayorov: You are absolutely right. Don't know how I missed it. Will update the PR. Well, just in case...

sarahmonster:
Since this fix is due to be released in core shortly, @iamdmitrymayorov are you able to update this PR for us? Might be best to create as a new PR, since there are some conflicts here.

iamdmitrymayorov:
@sarahmonster Absolutely, I'll update it shortly.

iamdmitrymayorov:
Closing this in favour of #884

PR #814, opened 2015-08-29
Set skip link focus element's outline to 0 programmatically

### LINKED ISSUE #543: Bug: skip-link-focus-fix.js causes unexpected outline on elements in Chrome

davidakennedy:
* Not doing so caused some elements to receive an unwanted `outline`.
* Setting this programmatically in the JavaScript avoids any additional CSS and keeps the changes contained to the script.
* It's okay to set the `outline` to 0 here since there's usually no need for a focus outline on an element that only receives focus programmatically.

Fixes #543

### DIFF
--- a/js/skip-link-focus-fix.js
+++ b/js/skip-link-focus-fix.js
@@ -24,6 +24,7 @@
 			if ( element ) {
 				if ( ! ( /^(?:a|select|input|button|textarea)$/i.test( element.tagName ) ) ) {
 					element.tabIndex = -1;
+					element.style.outline = '0px';
 				}
 
 				element.focus();


obenland:
It feels like a a style that should probably be in CSS, no?

samikeijonen:
Also noticed this "issue" several times and been thinking how to solve it. I like @davidakennedy idea. That  remove nasty blue border for skipped element. Not sure if you can do that using CSS because you don't know the element's ID.

obenland:
True, makes more sense now to me. `[id][tabindex="-1"] {}` could come reasonably close though.

davidakennedy:
Yeah, we could do it with CSS, for sure. The reason I went with the way I did is that the way the JavaScript is, the `outline` would show up on other anchor links clicked and focused on because of this script. So removing the `outline` with the JavaScript removes it on other anchor links beside just the `#content` one alone.

Does that make sense?

davidakennedy:
@obenland Bump. See my last comment. What do you think? How can we move this forward in some way?

obenland:
> the outline would show up on other anchor links clicked and focused on because of this script

If the script has an effect non-skip links, maybe we should review that script then?

davidakennedy:
> If the script has an effect non-skip links, maybe we should review that script then?

Well, maybe. Then issue that the script fixes, the fact that some browsers cannot programmatically move focus when an anchor link is clicked, effects not just traditional skip links, but any anchor link a theme author or user would create. So it's good that it works like it does.

My thinking here was also, let's remove the nastiness of the outline created by the script, but not remove the outline on everything, which is what `[tabindex="-1"]:focus { outline: 0; }` would do. There may be some cases where that's needed. We don't want to get in the way there.

obenland:
How many anchors with an id and a tabindex of -1 are on our pages? `a[id][tabindex="-1"]:focus` is not specific enough?

davidakennedy:
Okay, gotcha. I see what you're saying. So:

```css
#content[tabindex="-1"]:focus {
	outline: 0;
}
```

Putting that in `style.css` would solve the issue for our specific skip link, but leave the outline working on other anchors in case developers need that. I would be good with this solution, and happy to update this PR if everyone else is on board.

PR #816, opened 2015-09-05
Fix for the header color issue

carl-alberto:
This is in regards with the issue being discussed in here:
https://github.com/Automattic/_s/pull/785#issuecomment-136523557

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,14 +1,22 @@
 Thanks for contributing to `_s` (Underscores) — you rock!
 
-## _s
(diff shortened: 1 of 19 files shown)


iamtakashi:
The patch looks like including other changes but +1 to change the selector to be `.site-title a` for the color. That's something I had to change with themes I've worked on.

sarahmonster:
@carl-alberto would you be able to submit this as a new PR, or squash your commits into one?

carl-alberto:
@sarahmonster Sorry for the multiple PRs, I think this is the same PR that we are working with @davidakennedy I have the PR here https://github.com/Automattic/_s/pull/785/commits if it is a duplicate copy? can I close the duplicate PRs? thanks!

PR #817, opened 2015-09-09
Fix admin menu rollover issue w/ WP 4.3 & Chrome

StephenWidom:
With a simple addition to functions.php, we're able to fix the issue experienced when hovering over links in the admin sidebar when using the latest version of WordPress and Google Chrome.

http://wptavern.com/a-bug-in-chrome-45-causes-wordpress-admin-menu-to-break

### DIFF
--- a/functions.php
+++ b/functions.php
@@ -150,3 +150,12 @@ require get_template_directory() . '/inc/customizer.php';
  * Load Jetpack compatibility file.
  */
 require get_template_directory() . '/inc/jetpack.php';
+
+/**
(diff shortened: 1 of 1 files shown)


davidakennedy:
Hey @StephenWidom!

Thanks for the issue and pull request. Since this isn't caused by _s in any way, we'll wait for the Chrome development team to fix the issue. Until then, users can take the steps outlined in the WP Tavern post as a temporary solution.

Cheers!

StephenWidom:
That makes sense. Thanks for the timely response!

PR #819, opened 2015-09-16
Update contributor docs

### LINKED ISSUE #615: Sass and style.css: We need a contributor workflow.
It occurs to me that I am utterly confused with the proper workflow for patches and pull requests now that Sass is in place. Some PRs alter `style.css` while some PRs alter the Sass alternatives. @karmatosed and I had a little chat and it appears that we need to update our contributor guidelines so that people are submitting PRs in the right way and also primary _s collaborators know what to look for when PRs come in. @obenland, I see https://github.com/Automattic/_s/commit/b76f394a79b64075a0e82e7642b73b9c7d68ce27 only includes Sass stuff but no `style.css` stuff. And https://github.com/Automattic/_s/pull/605, which is a good fix, only includes `style.css` stuff but no Sass stuff. Very, very confused here, and feeling insecure about parity between Sass and the style file for the non-Sass folk.

How should we be handling this moving forward?

sarahmonster:
I've split this into two sections: submitting issues and contributing code, since those are two different actions. 

This fixes #615 by mentioning that both CSS and SCSS source files should be updated when submitting a PR.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,14 +1,22 @@
 Thanks for contributing to `_s` (Underscores) — you rock!
 
-## _s
-Before getting started, make sure your issue has not been discussed earlier. You can search for existing tickets [here](https://github.com/Automattic/_s/search).
+## Submitting issues
+Before submitting your issue, make sure it has not been discussed earlier. You can search for existing tickets [here](https://github.com/Automattic/_s/search).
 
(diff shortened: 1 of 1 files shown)


obenland:
Feel free to just push small changes like these right away, unless you feel it might be controversial or warrants further discussion. We can always reopen or revert if someone has objections.

PR #820, opened 2015-09-16
Remove hfeed from header.php, add to body_class filter. See #740

sixhours:
A patch, should we choose to go this route:

* Removes `hfeed` class from header.php
* Move application of the `hfeed` class into a conditional hooked to `body_class`, so it only applies to non-singular pages.

### DIFF
--- a/header.php
+++ b/header.php
@@ -21,7 +21,7 @@
 </head>
 
 <body <?php body_class(); ?>>
-<div id="page" class="hfeed site">
+<div id="page" class="site">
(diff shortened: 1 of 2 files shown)


dshanske:
I should have suggested it that way. Coincidentally, that is how I proposed it be added to core in https://core.trac.wordpress.org/ticket/30783

iamtakashi:
I think this is the right thing to do. I'm thinking to implement this in Twenty Sixteen also.

PR #824, opened 2015-09-18
restrict attached image width to 100%

aptarmy:
In underscores starter theme, we have set $GLOBALS['content_width'] but haven't yet set CSS rule which is recommended by WordPress official site.


alaczek:
We do set the max-width here: https://github.com/Automattic/_s/blob/master/style.css#L409

PR #826, opened 2015-09-21
Only output tagline markup when tagline exists or is_customize_preview

### LINKED ISSUE #825: Stray paragraph tag when tagline is empty
This is a problem that spans beyond _s, but it can be solved here:

If a user leaves the Tagline field empty, _s outputs an empty `<p></p>` in the markup. This is sub-optimal.

A quick search of the codex shows the current implementation of the tagline code in header.php is correct, but imo it is not good form to leave stray empty tags when content is not provided by the user.

Ideally there would be some sort of test to see if `get_bloginfo( 'description' );` is empty before the paragraph tags and function call are triggered. However, in the present state that results in the Tagline preview not working in Customizer if the user originally left the field blank.

tywayne:
Fixes #825 

As of WP 4.0 we have `is_customize_preview` which will return true if site is being previewed in the customizer. This allow us to conditionally output the `<p></p>` for the tagline, and it will still work in the customizer preview. 

This PR mimics [how it is handled in Twenty Sixteen](https://github.com/WordPress/twentysixteen/blob/master/header.php#L35-38).

### DIFF
--- a/header.php
+++ b/header.php
@@ -30,8 +30,12 @@
 				<h1 class="site-title"><a href="<?php echo esc_url( home_url( '/' ) ); ?>" rel="home"><?php bloginfo( 'name' ); ?></a></h1>
 			<?php else : ?>
 				<p class="site-title"><a href="<?php echo esc_url( home_url( '/' ) ); ?>" rel="home"><?php bloginfo( 'name' ); ?></a></p>
+			<?php endif;
+
+			$description = get_bloginfo( 'description', 'display' );
+			if ( $description || is_customize_preview() ) : ?>
+				<p class="site-description"><?php echo $description; ?></p>
 			<?php endif; ?>
-			<p class="site-description"><?php bloginfo( 'description' ); ?></p>
 		</div><!-- .site-branding -->
 
 		<nav id="site-navigation" class="main-navigation" role="navigation">


header.php
```diff
@@ -30,8 +30,12 @@
 				<h1 class="site-title"><a href="<?php echo esc_url( home_url( '/' ) ); ?>" rel="home"><?php bloginfo( 'name' ); ?></a></h1>
 			<?php else : ?>
 				<p class="site-title"><a href="<?php echo esc_url( home_url( '/' ) ); ?>" rel="home"><?php bloginfo( 'name' ); ?></a></p>
+			<?php endif;
+
+			$description = get_bloginfo( 'description', 'display' );
+			if ( $description || is_customize_preview() ) : ?>
+				<p class="site-description"><?php echo $description; ?></p>
```
philiparthurmoore: Would `<?php echo esc_html( $description ); ?>` be better? Having trouble keeping up with all the changing opinions around escaping, not escaping enough, and escaping too much 'round these parts.

FWIW here's what goes into all my themes:

```php
								<?php
								$description = get_bloginfo( 'description', 'display' );
								if ( ! empty( $description ) || is_customize_preview() ) : ?>
									<p class="site-description"><?php echo esc_html( $description ); ?></p>
								<?php endif; ?>
```

Pretty much the same as what you're proposing.

tywayne:
I personally don't think we need to escape here, as we're passing `'display'` as the second parameter to `get_bloginfo`.

If you compare the `bloginfo()` and `get_bloginfo()` functions, you'll see that even `bloginfo()` simply echos `get_bloginfo()` with `display` set as the second parameter.

That said, I'll still gladly add it if desired - clearly Travis CI doesn't like it not escaped :)

obenland:
> I personally don't think we need to escape here, as we're passing `'display'` as the second parameter to `get_bloginfo`.

Agreed.

philiparthurmoore:
Here's what I see in Core:

```php
        if ( 'display' == $filter ) {
                if ( $url ) {
                        /**
                         * Filter the URL returned by get_bloginfo().
                         *
                         * @since 2.0.5
                         *
                         * @param mixed $output The URL returned by bloginfo().
                         * @param mixed $show   Type of information requested.
                         */
                        $output = apply_filters( 'bloginfo_url', $output, $show );
                } else {
                        /**
                         * Filter the site information returned by get_bloginfo().
                         *
                         * @since 0.71
                         *
                         * @param mixed $output The requested non-URL site information.
                         * @param mixed $show   Type of information requested.
                         */
                        $output = apply_filters( 'bloginfo', $output, $show );
                }
        }
```

Does this not mean that someone can output malicious code via filtering `bloginfo`? I'm having trouble understanding why escaping wouldn't be a good practice here since it does not appear to be done elsewhere.

tywayne:
By default, core uses the `bloginfo` filter to run it through `wptexturize`, `convert_chars`, and `esc_html`. This is found in `wp-includes/default-filters.php`.

```php
// Format strings for display.
foreach ( array( 'comment_author', 'term_name', 'link_name', 'link_description', 'link_notes', 'bloginfo', 'wp_title', 'widget_title' ) as $filter ) {
	add_filter( $filter, 'wptexturize'   );
	add_filter( $filter, 'convert_chars' );
	add_filter( $filter, 'esc_html'      );
}
```

philiparthurmoore:
> By default, core uses the...

Lovely. Thanks for that very helpful information.

PR #827, opened 2015-09-22
Introduces singular.php and removes single.php and page.php

RajeebTheGreat:
Uses new singular.php which encourages DRY code.


obenland:
We've had a long discussion about the sense or nonsense of keeping template files dry when we worked on Twenty Thirteen. Back then it was about having just one `archive.php` vs. a template file for tags, categories, dates, etc. The core team's opinion was that it would contradict the concept of a template hierarchy, if we were to recreate specific cases within a template file. While initially I had a different opinion, I now understand where they were coming from and support that point of view.

If we wanted to be dry, or reduce the amount of template files in `_s`, we could just put everything in `index.php` and be done with it. But that's not what we want to do. Just because we can now use a `singular.php` template doesn't mean that we *have* to, or that it always makes sense to do so. As a matter of fact, I think the only time it does make sense is when content, regardless of type, will always be displayed in the same way. Which is not the case in `_s`.

@RajeebTheGreat, thanks for opening the PR and giving us the opportunity to consider it and form an opinion.
/cc @iamtakashi, @karmatosed.

karmatosed:
" As a matter of fact, I think the only time it does make sense is when content, regardless of type, will always be displayed in the same way. "

This to me seems like a solid base to make this decision. As a result it doesn't make sense for _s.

RajeebTheGreat:
Thanks for the explanation. 
Cheers!

PR #831, opened 2015-10-13
Add theme support for Responsive Videos.

thomasguillot:


obenland:
Can we keep the function documentation and change the existing formatting to closer resemble core standards?
https://make.wordpress.org/core/handbook/best-practices/inline-documentation-standards/php/#docblock-formatting

grappler:
I don't think this is needed in a starter theme and everyone has their own preference of solutions. I see that this may be useful for Automattic projects. (From a non-Automattic contributor perspective)

obenland:
It's a Jetpack feature, so it's only executed if Jetpack is active.

PR #841, opened 2015-10-29
Fix failing build in Travis

davidakennedy:
* See: https://travis-ci.org/Automattic/_s
* Build failed because of empty lines are not allowed in multi-line function.


obenland:
Feel free to just straight commit these types of changes in the future

davidakennedy:
Yeah, I thought about that, but wanted to make sure the build actually passed.

PR #845, opened 2015-11-05
Apply Text input CSS to Tel input as well

ardalann:
Tel is a html5 valid input type, and it should be styled similar to Text and Email inputs.


obenland:
See #280, #606, #613, and #614,

PR #853, opened 2015-11-11
Only generate pingback url header tag when relevant.

jrfnl:


karmatosed:
It's really helpful to get more information as to why you are suggesting something. For example, this isn't linked to an issue to help us see what you are helping to resolve. Could you please expand a little? It's always great to have a description in any PR.

jrfnl:
@karmatosed Thanks for letting me know, I'll expand. I wrongly assumed the code was self-explanatory.

1. The principle of pingbacks is based on articles. Pingbacks are a special kind of comments and the article being 'pinged-back' has to be identifiable (which they're not on archive pages and the like).
2. WP only registers a ping if it can identify the article which is supposed to have been 'pung'. See: https://developer.wordpress.org/reference/classes/wp_xmlrpc_server/pingback_ping/
3. Pingbacks, like comments, can be disabled on a per article basis.

Therefore, having the pingback url auto-discovery header in place, only makes sense on singular pages where pings are open.

Resulting in:
```php
if ( is_singular() && pings_open( get_queried_object() ) )
```


For further information:
* https://en.support.wordpress.com/comments/pingbacks/
* http://codex.wordpress.org/Introduction_to_Blogging#Pingbacks
* http://www.hixie.ch/specs/pingback/pingback
* http://www.w3.org/wiki/Pingback
* https://en.wikipedia.org/wiki/Pingback

obenland:
Adding logic in the header would make me want to move that into a `wp_head` callback. When `is_singular()` is true the post should be global and we probably wouldn't need to call `get_queried_object()`.

jrfnl:
@obenland Makes sense to me. Shall I adjust the PR & move the complete snippet to a function in `inc/extras/` ?

jrfnl:
Rebased & adjusted

philiparthurmoore:
I like this. Needs testing but it makes sense.

PR #855, opened 2015-11-11
Fix failing travis build

jrfnl:
Ignore WPCS output escape warning for description as it *is* already correctly escaped.

Fixes failing travis build.


davidakennedy:
Thanks for the fix @jrfnl! :smile:

jrfnl:
You're welcome.

PR #857, opened 2015-11-11
Do not show the outline on the skip link target container

### LINKED ISSUE #543: Bug: skip-link-focus-fix.js causes unexpected outline on elements in Chrome

davidakennedy:
* Putting it in style.css solves the issue for our specific skip link, but leaves the outline working on other anchors in case developers need that.
* For discussion, see: https://github.com/Automattic/_s/pull/814
* Related: https://github.com/Automattic/_s/pull/755 and https://github.com/Automattic/_s/pull/604

Fixes #543

### DIFF
--- a/sass/modules/_accessibility.scss
+++ b/sass/modules/_accessibility.scss
@@ -25,3 +25,8 @@
 		z-index: 100000; /* Above WP toolbar. */
 	}
 }
+
+/* Do not show the outline on the skip link target. */
+#content[tabindex="-1"]:focus {
+	outline: 0;
+}
--- a/style.css
+++ b/style.css
@@ -664,6 +664,11 @@ a:active {
 	z-index: 100000; /* Above WP toolbar. */
 }
 
+/* Do not show the outline on the skip link target. */
+#content[tabindex="-1"]:focus {
+	outline: 0;
+}
+
 /*--------------------------------------------------------------
 # Alignments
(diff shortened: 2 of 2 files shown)


mtomas7:
Should we also reposition the window on jump-to-anchor to account for admin bar like in:

https://github.com/WordPress/twentysixteen/commit/8d6241fc970b50e635e80faed8111a5b7b8b90c5

davidakennedy:
@mtomas7 Maybe? :smile: If you're game, submit a PR and we'll take a look. If you don't get to it, I'll see if I can't when I have time.

mtomas7:
Unfortunately I'm not a very good player yet, so will wait for you ;)

startover909:
I really hate to revive this once again, but coming across this myself on a recent project, I believe the CSS selector:
`#content[tabindex="-1"]:focus`
lacks a space between #content and []. Without the space this style does not work (tested in Chrome). Maybe a typo here?

davidakennedy:
@startover909 That's not a typo, unless I'm misunderstanding something. It scopes it exactly to the `#content` div when it has that attribute and focus. Removing it, or adding a space brings the focus indicator back for me in Chrome.

PR #858, opened 2015-11-11
Conditionally include Jetpack compatibility file.

jrfnl:
Lean loading - including files is _expensive_, so only include the file if it actually makes sense to do so.

### DIFF
--- a/functions.php
+++ b/functions.php
@@ -149,4 +149,6 @@ require get_template_directory() . '/inc/customizer.php';
 /**
  * Load Jetpack compatibility file.
  */
-require get_template_directory() . '/inc/jetpack.php';
+if ( defined( 'JETPACK__VERSION' ) ) {
+	require get_template_directory() . '/inc/jetpack.php';
+}


karmatosed:
Thank you for suggesting this, however in this case JetPack inclusion is up to author so lets not add this.

jrfnl:
@karmatosed While I agree that inclusion of JetPack support is up to the author, that doesn't take anything away from the validity of this pull request.

JetPack is a plugin after all and may be deactivated. If it is, it is unnecessary overhead to load the file and add the hook-ins. Lean loading should always be encouraged IMHO.

In other words, if the author includes JetPack support, this is the better way to do so. If the author chooses not to, they can/should remove the whole block.

I urge you to reconsider.

karmatosed:
@jrfnl thanks for your considered reply. I totally understand what you are saying. In this case however I have to stick by closing this. I have actually asked for a second opinion before closing and this was the outcome.

obenland:
I think we should reconsider. Obviously our response can always be "this is a starter theme, do what you want with it", that's easy. But I can see @jrfnl's point. The chance that Jetpack is not enabled is greater than that it is, checking for a constant would be a fairly cheap way to account for that.

davidakennedy:
Thanks for another pull request @jrfnl!

I see the point here, but think it's overkill in this case. But I'm open to being swayed.

A few things to keep in mind:

* In the environment we (Automattic) use _s on, WordPress.com, Jetpack is always active. So at least for us, it's a moot point.
* I wonder if checking for that constant in Jetpack on WordPress.com would be okay, since that version differs some from the .org version. That may not be an issue, but wanted to ask, just in case.

jrfnl:
> In the environment we (Automattic) use _s on, WordPress.com, Jetpack is always active. So at least for us, it's a moot point.

While I can appreciate that, the fact remains that _s is actively being promoted as a best practice starter theme via the http://underscores.me/ website and other means.
For a starter theme which is - apart from by Automattic - not going to be used (or even allowed?) on wp.com, the conditional loading is appropriate.

*__[playing devil's advocate...]__ Or - from a starter theme point of view - better yet, JetPack support should be removed completely as no other plugins are supported, so why make an exception for one ?*

So if you *do* make the exception for this one plugin - for very good and understandable reasons from an Automattic point of view -, at least implement support for it using lean loading best practices.

> I wonder if checking for that constant in Jetpack on WordPress.com would be okay, since that version differs some from the .org version. That may not be an issue, but wanted to ask, just in case.

I checked in the [JetPack GH repo](https://github.com/Automattic/jetpack/blob/master/jetpack.php) for an appropriate constant to check for. Not sure if the wp.com version of Jetpack is public - if so... where to find it ?
I'll gladly verify it if I have access.

obenland:
The actual version shouldn't make a difference as long as the constant is defined.

mendezcode:
I can see the motive & reasoning of adding the check, but as it stands currently, there are only 2 functions and 1 action hook defined in `inc/jetpack.php`, which has very low impact from a performance standpoint.

Underscores is a starter theme, which means things are meant to be modified and changed to suit the needs of the target theme implementation.

Adding the check makes sense if seen from another perspective, and it's also minimal. If the `inc/jetpack.php` file would be bigger then I'd vote in favor of it, but if the developer does not want jetpack integration, it's just a matter of deleting one line and removing the file.

jrfnl:
@mendezcode Sorry, but I think you are missing the point - it's about leading by setting a good example and using best practices.

Unfortunately not all theme developers have a proper understanding of WP let alone PHP. So we can't presume people are as knowledgeable as we are.

philiparthurmoore:
@jrfnl Can you clarify this?

> Lean loading - including files is expensive, so only include the file if it actually makes sense to do so.

How expensive? It'd help me understand the underlying issue more.

jrfnl:
@philiparthurmoore Can't remember exactly where I saw the stats, will try to find them for you, but if I remember correctly, 20% of processing time of an average page load for a WP page is just and only caused by includes/requires.

Generally speaking anything which needs access to the file system - and that includes file loads - are, relatively speaking, "slow" operations in PHP.

philiparthurmoore:
> Underscores is a starter theme, which means things are meant to be modified and changed to suit the needs of the target theme implementation.

“_s is a starter theme” is up there with “you can always fork it” for sentences that have no meaning or substance for specific issues. 

The pull request itself is an appeal to add or remove code to improve the starter theme. By this point “starter theme” has a ton of definitions—I know, I wrote one—so using it to shut down introspection on whether or not something should be added or removed is on the intellectually lazy side.

If there are clear stats that doing this would improve _s’s load times, it's a good move. 

If it's simply a matter of theory and ”best practices” without any real performance impact then it's indeed probably up to subjective viewpoints.

davidakennedy:
> If there are clear stats that doing this would improve _s’s load times, it's a good move.

I agree on this. I'd definitely be open to making the change if I can better understand the impact. Admittedly, right now I don't have a good grasp of the difference.

davidakennedy:
I wanted to come back to this since we never came to a conclusion. It does seem like a good idea overall, so thanks to everyone for the discussion and thoughts.

I'm double checking on the `JETPACK__VERSION` constant to make sure it can also be used on WordPress.com. If it can be, I'd say let's merge it. If it can't, we should be able to find something else usable for that check.

kraftbj:
`JETPACK__VERSION` is set to `wpcom` on WordPress.com. If that file should _not_ be loaded on WP.com, it would need to check for that. If the file should be loaded on WP.com, then checking for that constant is fine.

PR #734, opened 2015-04-23
Update non-https WordPress URLs.

lukemcdonald:

PR #863, opened 2015-11-13
Update non-https WordPress URLs. #734

limestreet:
Hi, there is a patch for #734. It looks like last 3 changes in urls.


limestreet:
I'm not sure about changing http to https in _s.pot file header probably this file is auto generated. So maybe second patch will be better...

limestreet:
This PR was little bit messed up so I submitted new one #864

PR #864, opened 2015-11-13
Update non-https WordPress URLs without pot file

limestreet:
There is a patch for the last two instances of http. The another one is in _s.pot file header but I think pot file is auto generated, so every changes will be overwritten. It's for ticket #734

### DIFF
--- a/inc/custom-header.php
+++ b/inc/custom-header.php
@@ -10,7 +10,7 @@
 	</a>
 	<?php endif; // End header image check. ?>
  *
- * @link http://codex.wordpress.org/Custom_Headers
+ * @link https://codex.wordpress.org/Custom_Headers
  *
  * @package _s
  */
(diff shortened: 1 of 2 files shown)


emiluzelac:
To reflect all recant link changes, I would actually suggest the following instead:

- https://developer.wordpress.org/themes/functionality/custom-headers/

Could not find RTL within https://developer.wordpress.org/themes/

sarahmonster:
These links aren't present in the POT file since they're just code comments—not something that's translated. For reference, the POT file isn't automatically generated. I usually just edit it manually to make changes where required. 

I'm on board with changing the custom header link to the new documentation. @limestreet, would you be able to update this PR, and then we can merge?

limestreet:
Hi @sarahmonster - sure I'll update this soon :)

PR #867, opened 2015-11-21
Move aside from widgets to the entire sidebar.

WPAddiction:
It appears that the entire sidebar should be one large aside instead of making each widget it's own aside. At least that's how I understand it when reading w3. It would make sense to make each widget it's own aside if not all of the widgets were asides, but otherwise the entire sidebar should be considered one aside.

Plus I think this would make more sense for accessibility including widget titles being h2s because if the entire sidebar had a title that would be the h1.

https://www.w3.org/wiki/HTML/Elements/aside

### DIFF
--- a/functions.php
+++ b/functions.php
@@ -102,8 +102,8 @@ function _s_widgets_init() {
 		'name'          => esc_html__( 'Sidebar', '_s' ),
 		'id'            => 'sidebar-1',
 		'description'   => '',
-		'before_widget' => '<aside id="%1$s" class="widget %2$s">',
-		'after_widget'  => '</aside>',
+		'before_widget' => '<div id="%1$s" class="widget %2$s">',
+		'after_widget'  => '</div>',
 		'before_title'  => '<h2 class="widget-title">',
 		'after_title'   => '</h2>',
 	) );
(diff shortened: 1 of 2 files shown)


WPAddiction:
I did a little poking around after my initial pull request and found a discussion about this very thing for the new default theme Twenty Sixteen. That's where I got the idea to change the individual widget wrappers from div to section, which I think does make sense.

https://github.com/WordPress/twentysixteen/issues/355

karmatosed:
@iamtakashi and @davidakennedy looping you in here to get your opinion on whether this should go into _s.

iamtakashi:
It makes sense to change each widget to `section` to reduce number of landmarks, and I believe the widget area should then be `aside`.

So, yes, +1 from me.

davidakennedy:
I agree with Takashi here. This makes sense to do. :+1: from me.

iamtakashi:
It's awesome to see the effort made in Twenty Sixteen is lopping back to _s. Thanks @WPAddiction.

pstonier:
Glad to see this in here. It's been a point of confusion for me when creating themes from _s and I go to set the sidebar as an aside, then realize the widgets themselves are asides.

PR #870, opened 2015-11-25
Update posted_on template tag function name.

helloguna1:
The name posted on is a little misleading as it also contains other
meta information of the entry such as “by”.

I have changed it to _s_entry_meta which follows the _s_entry_footer naming.

### DIFF
--- a/inc/template-tags.php
+++ b/inc/template-tags.php
@@ -7,11 +7,11 @@
  * @package _s
  */
 
-if ( ! function_exists( '_s_posted_on' ) ) :
+if ( ! function_exists( '_s_entry_meta' ) ) :
 /**
  * Prints HTML with meta information for the current post-date/time and author.
  */
-function _s_posted_on() {
+function _s_entry_meta() {
 	$time_string = '<time class="entry-date published updated" datetime="%1$s">%2$s</time>';
 	if ( get_the_time( 'U' ) !== get_the_modified_time( 'U' ) ) {
 		$time_string = '<time class="entry-date published" datetime="%1$s">%2$s</time><time class="updated" datetime="%3$s">%4$s</time>';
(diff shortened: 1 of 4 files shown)


karmatosed:
I'm not sure entry meta is good for this as that infers other content too. I'd vote to note change this and keep as is.

helloguna1:
I took a look at the default themes and only Twentyfourteen uses posted_on(). Themes Twentythirteen, fifteen and sixteen use entry_meta() (not really sure why they changed it back in fourteen only to change it again in the next 2 themes).

The name posted_on() is a little too specific while entry_meta() is more meaningful as categorize a group of data.

sarahmonster:
Looking at TwentySixteen as an example, it has sensibly-named functions in this file (https://github.com/WordPress/twentysixteen/blob/master/inc/template-tags.php) that follow the same overall naming convention and are indicative of the content they output—`twentysixteen_entry_meta()`, `twentysixteen_entry_date()`, etc. This is a pretty sane & logical approach, but it doesn't really fit with how _s is using these functions, or how theme authors are likely to rearrange the data within the functions.

`twentysixteen_entry_meta()` prints data about categories, tags, etc, so renaming the function that contains the date and author here in _s doesn't feel like the right choice. 

Here's what we have right now:

`_s_posted_on` appears at the top of the post and contains the date posted and author byline
`_s_entry_footer` appears at the bottom of the post and contains the categories, tags, comment link, and edit link.

I think these might do with considered renaming, given that theme authors are likely to shuffle elements around within the two functions, to encourage good practises here, but I don't think using `_s_entry_meta` is the best solution. I'm going to close this PR—if you'd like to discuss the issue further, please feel free to open a new issue and we can discuss in more detail. 

Cheers!

PR #880, opened 2015-12-04
Adjust travis script to include testing against PHP 7 and HHVM.

jrfnl:
PHP 7 has officially been released (since yesterday), so should be tested against. HHVM is not a bad idea to keep in mind either.
Also noted that WP4.3 wasn't included yet in the travis run.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,6 +10,8 @@ language: php
 
 # Declare versions of PHP to use. Use one decimal max.
 php:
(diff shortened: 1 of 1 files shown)


obenland:
That has @philiparthurmoore written on it.

philiparthurmoore:
I'll look at this tomorrow. :)

philiparthurmoore:
Let's close this and handle it all at the same time in #881.

PR #882, opened 2015-12-06
Use consistent syntax for control structures

byjml:
Changed the opening brace to ':' (or colon) and the closing brace to `endif;`

### DIFF
--- a/template-parts/content.php
+++ b/template-parts/content.php
@@ -12,11 +12,11 @@
 <article id="post-<?php the_ID(); ?>" <?php post_class(); ?>>
 	<header class="entry-header">
 		<?php
-			if ( is_single() ) {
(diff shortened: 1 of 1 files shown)


grappler:
The reason for using brackets here is that this is one block of PHP and not switching with HTML Related: #343

:-1:

obenland:
This is a template file however

byjml:
@grappler Isn't it a different case for template files?

philiparthurmoore:
@grappler You still against this?
@obenland What do you think?

grappler:
After relooking at it is does not look a bit better in this format.

sheabunge:
I'm not a fan of the extended structure, especially when not using it with HTML. It's less clear where the structure begins and ends

PR #883, opened 2015-12-07
Remove Normalize right and left padding from figure element.

jwogrady:
Normalize resets the `figure` tag as follows:

    figure {
	    margin: 1em 40px;
    }

The caused the [wptest data's full size images with captions](http://wptest.io/demo/category/images/) to overlap my Susy grid in Safari.  Removing the right and left padding from the figure elements fixed the issue.  Now the 1200px wide images are fully contained within the grid and responsive across all views.

### DIFF
--- a/sass/typography/_copy.scss
+++ b/sass/typography/_copy.scss
@@ -43,3 +43,7 @@ mark, ins {
 big {
 	font-size: 125%;
 }
+
+figure {
+	margin: 1em 0px;
+}


obenland:
I'd prefer to keep normalize as is so we can just keep dropping it in when it gets updates. Conflicts with other libraries it should probably be mitigated upstream.

WPAddiction:
It's not an edit to _normalize.scss @obenland, so I don't see a problem.

I personally add something similar to my themes, I just have it in _elements.scss.

Also, if this gets approved, don't forget to add the same thing to style.css.

karmatosed:
I'm with @obenland on keeping as is for now.

PR #884, opened 2015-12-07
Introducing comment navigation template tags.

dmtrmrv:
This PR replaces the #811. Since the comment navigation template tags are in 4.4 (https://core.trac.wordpress.org/ticket/30589) we can now use them, providing the fallback in template-tags.php for backward compatibility.

I still have one concern though:

```
echo get_the_comments_navigation( $args );  // WPCS: XSS OK.
```

Added a `WPCS: XSS OK` to hide codesniffer errors, because it seems that all translatable strings are escaped in `get_the_comments_navigation()` and `_navigation_markup()`. Would be great if someone could take a second look.

### DIFF
--- a/comments.php
+++ b/comments.php
@@ -45,17 +45,7 @@ if ( post_password_required() ) {
 			?>
 		</h2><!-- .comments-title -->
 
-		<?php if ( get_comment_pages_count() > 1 && get_option( 'page_comments' ) ) : // Are there comments to navigate through? ?>
-		<nav id="comment-nav-above" class="navigation comment-navigation">
-			<h2 class="screen-reader-text"><?php esc_html_e( 'Comment navigation', '_s' ); ?></h2>
-			<div class="nav-links">
-
-				<div class="nav-previous"><?php previous_comments_link( esc_html__( 'Older Comments', '_s' ) ); ?></div>
-				<div class="nav-next"><?php next_comments_link( esc_html__( 'Newer Comments', '_s' ) ); ?></div>
-
-			</div><!-- .nav-links -->
-		</nav><!-- #comment-nav-above -->
-		<?php endif; // Check for comment navigation. ?>
+		<?php the_comments_navigation(); ?>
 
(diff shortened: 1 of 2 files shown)


philiparthurmoore:
This has languished. @iamdmitrymayorov, should we still look into this?

iamdmitrymayorov:
@philiparthurmoore Sure, I'd be happy to review/update the PR myself if the idea worth the merge.

samikeijonen:
Any news on this?

davidakennedy:
@dmtrmrv Thanks for the work here! I'd love to get this in ASAP. Any interest on refreshing this PR? Or can someone else continue?

inc/template-tags.php
```diff
@@ -77,6 +77,63 @@ function _s_entry_footer() {
 }
 endif;
 
+if ( ! function_exists( 'get_the_comments_navigation' ) ) :
```
davidakennedy: The fallback can be removed since 4.8 is out now, and this PR sat for awhile.

dmtrmrv:
Great news! Sure, I'll update the PR shortly.

dmtrmrv:
@davidakennedy, All done.

dmtrmrv:
@grappler Sure, just generated the .pot file with `grunt-wp-i18n` but it looks like it is changing the format of the .pot file. For example, it removes some portions of string declarations like:
```
#, php-format
#@ _s`
```
Is it normal? Or am I doing something wrong here? Thanks.

grappler:
@dmtrmrv `#@ _s` was introduced in #767 but there was no reason given why it was added. I have also tried generating the pot file with poedit and I get the same results.

FYI There is an NPM tasks that may be quicker than grunt `npm install -g node-wp-i18n` to install and  `wpi18n makepot --type=theme --domain-path=language` to run.

dmtrmrv:
@grappler Thanks for the clarification! I'll give that script a try.

PR #887, opened 2015-12-08
list all html5 input types. fixes #885

### LINKED ISSUE #885: HTML5 Input Type Styles Redux
*Yes. I've read #280, #606, #613, #614, and #845. I'm hoping we can reopen this discussion because it's 2015 and I've got some data points to share. I'm happy to put together a PR, but I thought I should ask before investing any additional work.*

## tl;dr
 It's 2015 and time to style all (or at least `number` and `tel`) HTML5 input types. Core, form plugins, and the internet at large are all going that direction.

## The primary issue
\_s only includes styling for _some_ HTML5 input types and excludes at least two common ones.

## Back story
I've been using some Automattic themes recently and [filed an issue on Sequential](https://wordpress.org/support/topic/no-support-for-inputtypetel?replies=8) about `input[type="tel"]` and `input[type="number"]` recently. I ran into the issue again with Sela and realized I needed to look upstream since it's obviously coming from a similar code base. That took me to [Automattic's Theme Pattern Library](https://github.com/Automattic/theme-pattern-library) and then back to \_s. (Six-degrees of \_s, anyone?)

Here's a simple Gravity Forms form using Sela (without my child theme styles to fix this all):

basic gravity forms form in Sela theme

Note "Phone" and "Total Number of Guests" have a different border color (among other things). When I look at why this is, it's pretty easy to directly track this back to the included list of HTML5 input types in \_s.

## Previous arguments against inclusion:
@obenland in [Aug 2013](https://github.com/Automattic/_s/pull/280#issuecomment-22466443):
> Let's stick with what comes with core for now.

@karmatosed in [Sept 2014](https://github.com/Automattic/_s/pull/614#issuecomment-55486328):
> My feelings were that we should support the common ones. Even when I was freelance I can't recall the last time tel was supported in something I did, I don't think it's that common.
> This all said, if we did include it we should include everyone - perhaps a slippery slope we may not want to go down yet?

## Data
As best I could in an hour+, I reviewed Core, Jetpack, Gravity Forms, Contact Form 7, Ninja Forms, and Formidable (all power tens of thousands of sites on the low end). You can see [my preliminary data](https://docs.google.com/spreadsheets/d/1amW5LiM72o5xTMXOpKYAV7QOAD2sA9HVGfknsw7bXBA/edit?usp=sharing). Please let me know if I can update.

### Results
- If we were to includes styles for all fields types that Core outputs on the front _or_ backend: text, email, password, number, url, and search. (I'm leaving our file for now.)
- If we were to includes styles for all fields types that at least two plugins include: text, email, number, password, url, tel. (Range and Date are only supported by CF7)
- If we were to includes styles for all fields types that Core *styles*: All valid HTML5 input types (see `/wp-admin/forms.css`)

## A few more observations
- These input types have [strong browser support](http://www.wufoo.com/html5/#types).
- Most (all?) of the HTML5 input types fallback to `input[type="text"]` in unsupported browsers, so graceful degradation means we can use these *now*.
- With the rise of mobile, both `number` and `tel` trigger different keyboards in most mobile OS's, greatly increasing the [benefits of using these input types](http://www.smashingmagazine.com/2015/05/form-inputs-browser-support-issue/).
- [Google recommends using these](https://developers.google.com/web/fundamentals/design-and-ui/input/forms/choose-the-best-input-type?hl=en).

## Conclusion
I think both of the reasons for _not_ doing this are gone:
- Core anticipates these types' usage and includes styles for them
- `tel` and `number` are both fairly widely used

There are good reasons to use these input types and not many reasons not to. \_s tries to be modern and follow best practices (sometimes too much, in the case of things like HTML5 heading structure), but I think this is one place where clearly \_s is lagging behind modern developer practices.

### Moving forward?
The only question then is how to improve the styles. Do we include a list of all valid HTML5 inputs or do we style `input` and then apply additional styles to inputs that shouldn't be "text-y" in appearance (e.g. `input[type="button"]`)? (This was raised by @philiparthurmoore in [Sep 2014](https://github.com/Automattic/_s/pull/614#issuecomment-55486419).).

Based on what I've seen others do, I think listing them all—while slightly more verbose—is probably the way to go. I'm not super up on this, but I haven't seen anything to suggest that more input types will be added any time soon. (And then we can have this whole discussion all over again!)

mrwweb:
- This includes .css and .scss updates.
- I added a border for the `select` element to match other inputs.
- Can anyone explain the padding-left: 3px on [Line 488](https://github.com/Automattic/_s/blob/master/style.css#L488)? It seems to have come in the initial start-styles commit and I can't find any explanation or browser bugs that would explain it. The reason I ask is mostly because it requires listing out every input type twice, the second time solely for the purpose of padding. I'd love to just use the same padding for `textarea` but don't want to break anything. I'm happy to amend if that would be useful.

### DIFF
--- a/sass/forms/_fields.scss
+++ b/sass/forms/_fields.scss
@@ -3,6 +3,16 @@ input[type="email"],
 input[type="url"],
 input[type="password"],
 input[type="search"],
+input[type="number"],
+input[type="tel"],
+input[type="range"],
+input[type="date"],
+input[type="month"],
+input[type="week"],
+input[type="time"],
+input[type="datetime"],
+input[type="datetime-local"],
+input[type="color"],
 textarea {
 	color: $color__text-input;
 	border: 1px solid $color__border-input;
@@ -13,11 +23,25 @@ textarea {
 	}
 }
 
+select {
+	border: 1px solid #ccc;
+}
+
 input[type="text"],
 input[type="email"],
 input[type="url"],
 input[type="password"],
-input[type="search"] {
+input[type="search"],
+input[type="number"],
+input[type="tel"],
+input[type="range"],
+input[type="date"],
+input[type="month"],
+input[type="week"],
+input[type="time"],
+input[type="datetime"],
+input[type="datetime-local"],
+input[type="color"] {
 	padding: 3px;
 }
 
--- a/sass/forms/_fields.scss
+++ b/sass/forms/_fields.scss
@@ -24,7 +24,7 @@ textarea {
 }
 
 select {
-	border: 1px solid #ccc;
+	border: 1px solid $color__border-input;
 }
 
 input[type="text"],
--- a/style.css
+++ b/style.css
@@ -35,7 +35,7 @@ Nicolas Gallagher and Jonathan Neal http://necolas.github.com/normalize.css/
 # Clearings
 # Widgets
 # Content
-    ## Posts and pages
+	## Posts and pages
 	## Asides
 	## Comments
 # Infinite scroll
@@ -461,10 +461,24 @@ input[type="email"],
 input[type="url"],
 input[type="password"],
 input[type="search"],
+input[type="number"],
+input[type="tel"],
+input[type="range"],
+input[type="date"],
+input[type="month"],
+input[type="week"],
+input[type="time"],
+input[type="datetime"],
+input[type="datetime-local"],
+input[type="color"],
 textarea {
 	color: #666;
 	border: 1px solid #ccc;
 	border-radius: 3px;
+}
+
+select {
+	border: 1px solid #ccc;
 }
 
 input[type="text"]:focus,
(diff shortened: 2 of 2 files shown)


mrwweb:
Variable is fixed. Thanks, @iliman.

Before this is finalized, is there a reason for `textarea`'s `padding-left: 3px` that we know about? If `textarea` can just have `padding: 3px`, we can cut about 15 lines out.

siliconforks:
For the textarea padding - maybe it was done that way because of a Firefox bug?

https://bugzilla.mozilla.org/show_bug.cgi?id=157846

mrwweb:
@siliconforks that probably is it!

Looking at the bug report and the [test case in Firefox](http://jsfiddle.net/af47whzf/1/), I'm tempted to say that the remaining issue (text scrolls "behind" top and bottom padding) in Firefox isn't all that bad, particularly with smaller padding values. The specific padding bug addressed directly by `padding-left: 3px` (right padding between the border and scrollbar) was fixed in early 2014, so I'd suggest applying `padding: 3px` to the `textarea`. Here's how that looks:

Chrome:
image

Firefox:
image

Anyone else want to weigh in?

And FWIW, I quickly browsed [sanitize.css](https://github.com/10up/sanitize.css/blob/master/sanitize.css), [normalize.css](https://github.com/necolas/normalize.css/blob/master/normalize.css), and [formalize.css](https://github.com/nathansmith/formalize/blob/master/assets/css/formalize.css). None include code to address this issue, and formalize actually gives `textarea`s the same padding as `input`s.

mrwweb:
If the `textarea` discussion is derailing things and this PR is otherwise ok, I'd suggest just accepting it and I submit the `textarea` change as a separate PR for further discussion.

mrwweb:
Seems like this has stalled out. I'd like to officially propose that this PR be merged as-is (if it's going to be merged) and that I submit a second pull request to deal with the `textarea` issues.

Does anything else need to be done to make this happen? I'm happy to tweak the existing PR quickly if needed.

PR #888, opened 2015-12-12
Slim down the travis script - remove superfluous checks.

### LINKED ISSUE #881: Travis script is highly inefficient
On a related note to PR #880, I should like to open the discussion on the travis script in general.

This would be a great setup if there were unit tests to run as they should pass against all sorts of possible scenarios.

However, as the only automated tests being run are a PHPCS check and a PHP lint, the current script is very inefficient and contains lots of redundancy:
* having a WP version matrix in the travis script is redundant as the results of PHPCS and lint checks will not change across WP versions.
* cloning WP is redundant as it is currently not needed nor used.
* linting against anything *__but__* *one* low (5.2) and *one* high PHP version (5.6/7.0) and possibly outliers (HHVM) is redundant.
* the same goes for the PHPCS check.

The current script takes a long time to run and does not comply with the fair use guidelines of Travis in its current form.
A much simpler version would be sufficient (until the time that unit tests will be/have been created).

jrfnl:
Follow up on #881 and #880.

What it will now do:
- Lint the php files against relevant PHP versions.
- Lint the js files once - the result won't change across PHP versions.
- Check against WPCS once - the result won't change across PHP versions.

What I have changed:
- Added linting against PHP 7 and HHVM, with HHVM being allowed to fail.
- Added js linting and style check per example from Twenty Sixteen.
- Moved to the faster container based environment for running travis.
- Script will no longer pull in PHPCS, WPCS and the JS linters in every build. Now they will only be pulled in when needed.
- Limited the clone depth for quicker cloning of external repositories.
- Removed the pulling in of WP and running builds against different WP versions as this wasn't used at all in the actual test scripts.

What could still be slimmed down:
- linting against PHP 5.2, 5.6, 7.0 and HHVM should be sufficient to catch any pertinent issues, so the intermittent PHP versions *could* be removed.

Closes #881.

------------
Edit: The js checks threw some errors. Fixed those in a separate commit included in this PR.

For the record:
* time of a travis run before: approx: __11__ minutes.
* time of a travis run after: approx: __1½__ minutes.


philiparthurmoore:
Can you squash this into a single commit?

jrfnl:
> Can you squash this into a single commit?

Of course, just felt that it would be more descriptive to have two :wink:

philiparthurmoore:
:smile: 

Thanks. Midnight in Hanoi. Probably won't get around to reading through this until tomorrow or Monday.

jrfnl:
No worries, squash done and expanded the commit message for prosperity.

philiparthurmoore:
Wonderful. Thanks so much.

.jshintignore
```diff
@@ -0,0 +1,2 @@
+js/skip-link-focus-fix.js
+**.min.js
```
philiparthurmoore: Why are you excluding `skip-link-focus-fix.js`? I do not think that `**.min.js` is necessary here.

.jscsrc
```diff
@@ -0,0 +1,7 @@
+{
+    "preset": "wordpress",
+    "fileExtensions": [ ".js" ],
+    "excludeFiles": [
+    	"js/skip-link-focus-fix.js"
```
philiparthurmoore: `skip-link-focus-fix.js` shouldn't be excluded.

.jshintrc
```diff
@@ -0,0 +1,26 @@
...
+	"browser": true,
+
+	"globals": {
+		"_": false,
+		"Backbone": false,
+		"jQuery": false,
+		"JSON": false,
+		"wp": false
+	}
+}
```
philiparthurmoore: We can pull this file in with `wget https://develop.svn.wordpress.org/trunk/.jshintrc`. No need to create disparity between _s and how Core does things. See [this](https://github.com/WordPress/twentysixteen/blob/master/.travis.yml#L75).

.travis.yml
```diff
@@ -4,47 +4,40 @@
 # For use with the Underscores WordPress theme.
 # @link https://github.com/Automattic/_s
 
+# Ditch sudo and use containers.
+# @link http://docs.travis-ci.com/user/migrating-from-legacy/#Why-migrate-to-container-based-infrastructure%3F
+# @link http://docs.travis-ci.com/user/workers/container-based-infrastructure/#Routing-your-build-to-container-based-infrastructure
+sudo: false
+
```
philiparthurmoore: Good.

.travis.yml
```diff
@@ -4,47 +4,40 @@
...
+        # Current $required_php_version for WordPress: 5.2.4
+        # aliased to 5.2.17
+        - php: '5.2'
+        # aliased to 5.3.29
+        - php: '5.3'
+        # aliased to a recent 5.4.x version
+        - php: '5.4'
+        # aliased to a recent 5.5.x version
+        - php: '5.5'
+          env: SNIFF=1
```
philiparthurmoore: Why'd you put `SNIFF=1` here? Arbitrary decision? It's fine, just curious.

.travis.yml
```diff
@@ -55,57 +48,46 @@ matrix:
...
-  # @link http://pear.php.net/package/PHP_CodeSniffer/
-  # -p flag: Show progress of the run.
-  # -s flag: Show sniff codes in all reports.
-  # -v flag: Print verbose output.
-  # -n flag: Do not print warnings. (shortcut for --warning-severity=0)
-  # --standard: Use WordPress as the standard.
-  # --extensions: Only sniff PHP files.
-  - $WP_DEVELOP_DIR/php-codesniffer/scripts/phpcs -p -s -v -n . --standard=./codesniffer.ruleset.xml --extensions=php
+    # Search for PHP syntax errors.
+    - find -L . -name '*.php' -print0 | xargs -0 -n 1 -P 4 php -l
```
philiparthurmoore: Is this the only line that'll get run against the matrix? Looks like everything else is relegated to a single sniff (which is fine).

philiparthurmoore:
Left a few notes.

.jscsrc
jrfnl: I excluded it as it looked like it came from an external source which does not use the same coding standards as WP.

Fixing the file up is easy enough, but would have to be done again if a newer version of `skip-link-focus.js` at some point would be used.

.jshintignore
jrfnl: > Why are you excluding skip-link-focus-fix.js?

See above.

> I do not think that **.min.js is necessary here.

Although `_s` does not supply minified files (may be it should), themes based on `_s` often will. Having a sane base for `.jshintignore` made sense when I set it up.

.jshintrc
jrfnl: The build failed because the file couldn't be pulled in. Would be great if we could get it to work though.

.travis.yml
jrfnl: Arbitrary decision. Lowest PHP version which is still supported by PHP itself.

.travis.yml
jrfnl: > Is this the only line that'll get run against the matrix?

True - In reality, we could even remove some PHP versions and only run this lint against PHP 5.2, 5.6, 7.0 and HHVM. I don't think the versions in between will throw different parse errors which aren't caught by running just those four.

> Looks like everything else is relegated to a single sniff (which is fine).

Yes, the results of everything else won't change across PHP versions.

jrfnl:
Responded to all ;-)

jrfnl:
Added a new commit which updates the PR based on the feedback.

Changes:
* Include skip-link-focus-fix.js in the js hint/lint tests & fix up the file.
* Pull in .jshintrc from WP SVN.
* Add PHPCompatibility Sniffs.
* Sync the ignore statements to always exclude .min.js files.
* Slim down the tested against PHP versions even more.

@philiparthurmoore Separate PR to quickly see the changes, will squash them into one commit once you've had a chance to look at it.

philiparthurmoore:
Thanks. Will give this another sweep tomorrow. Apologies for the slow (not slow, really, but not fast) review. Pre-Christmas work is crazy. Thanks for this. I'll have more feedback for you soon.

jrfnl:
Sounds good to me and no worries, I know the feeling.

jrfnl:
Oh - one more commit (will be squashed too), but as the js files have all changed, they need to be broken out of the cache. Fixed that in this last commit.

philiparthurmoore:
Thanks. We'll get this in sooner than later. Need a little more time to read through all this and make sure it's all good. Thanks for the patience and for making this better.

karmatosed:
@philiparthurmoore where are you on this? I ask as seems you were working through it.

philiparthurmoore:
Haven't had any time to look at it - if someone else wants to awesome. If
not, I probably won't be back from Lunar New Year break until next week to
get it in.

obenland:
I can take a look

philiparthurmoore:
Cool.

Issue #786, opened 2015-07-29
Unnecessary roles in header, nav, main and footer


11bits:
I checked the theme with https://validator.w3.org and according this tool:

- role="banner" and role="contentinfo" are unnecessary for elements that are not descendants of an article element or a section element. 
- Navigation role and main role are not necessary for elements nav and main.

davidakennedy:
Hey @11bits!

Thanks for creating this issue! This is a bit tricky, but after some research, this is where I'm landing as far as our approach in _s for this:

* These errors are definitely valid errors. The advice for explicitly adding landmark roles is changing. See this [post](http://alistapart.com/comments/wai-finding-with-aria-landmark-roles) and [comment](http://alistapart.com/comments/wai-finding-with-aria-landmark-roles#338902).
* I brought this to the WordPress Accessibility team for [other thoughts](https://wordpress.slack.com/archives/accessibility/p1438627360000954).
* Currently, _s supports IE9 and up.
* Internet Explorer is the only major browser that [doesn't support the semantic information via accessibility api](http://www.html5accessibility.com/). But there are bugs filed, so that will likely change in future versions. Maybe Edge? The support is better when the roles are there, especially with a wider range of assistive technology, like screen readers.

Given that we support IE9 and up, I'm inclined to keep the roles explicitly attached to the html5 elements they correspond with because that's better for the users we support with _s. So we'll let the errors in the validator go for now. For what it's worth, we had errors for having `role`s specified for awhile until the validators caught up.

This is an issue we should definitely revisit as support gets better and/or as we drop support for older browsers.

You may want to take the `role`s off for your _s-based project, depending on your browser and assistive technology support.

Thanks again! I'm closing this issue, but if anyone wants to discuss further, just comment, and we'll reopen.

PR #890, opened 2015-12-13
Remove default implicit ARIA semantics

### LINKED ISSUE #889: Remove default implicit ARIA semantics
>  Web developers should not set the ARIA role and aria-* attributes to values that match the default implicit ARIA semantics defined in the table.

Quoted from the [W3C working draft about ARIA in HTML](http://www.w3.org/TR/html-aria/#document-conformance-requirements-for-use-of-aria-attributes-in-html).

The default implicit ARIA semantics should not be used in the markup.

PR #890

benoitchantre:
The default implicit ARIA semantics should not be used in the markup.

Fixes #889

### DIFF
--- a/404.php
+++ b/404.php
@@ -10,7 +10,7 @@
 get_header(); ?>
 
 	<div id="primary" class="content-area">
-		<main id="main" class="site-main" role="main">
+		<main id="main" class="site-main">
 
 			<section class="error-404 not-found">
 				<header class="page-header">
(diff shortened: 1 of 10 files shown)


emiluzelac:
Yep, I was looking at that myself as well.

screen shot on 2015-12-20 at 13-55-48

samikeijonen:
Here is related issue #786.

emiluzelac:
Makes sense, thanks!

sixhours:
Closing as per discussion in #889

PR #893, opened 2015-12-18
Fixed SASS spacing for consistency.

MarkBatchelder:
Just added lines where there should have been one and broke some lines into multiple lines where they included more than one element. Really just added lines to SASS files to keep things consistent with the rest of the SASS and CSS.


obenland:
Removing new lines from the end of those files would introduce an inconsistency with the rest of `_s` files.

WPAddiction:
I put all of the blank lines at the end of files back. Obviously I didn't think they were necessary, :P but I can see what you're talking about.

karmatosed:
Can you look at the CSS output too/

WPAddiction:
Slowly reviewing things; actually finding a couple more things I could change for consistency and now matching it up with style.css.

WPAddiction:
There are two files that still don't match up, but they have nothing to do with the changes I've made. We could also change the order of a few elements in style.css to match the way the SASS is compiled because of nesting. But the only other differences are the fault of the SASS compiler.

WPAddiction:
To clarify, I just added spaces between selectors that needed them and listed selectors on separate lines instead of a single line.

Before
`blockquote, q {`

After
`blockquote,`
 `q {`

Also: https://github.com/Automattic/_s/commit/9a08ca0bfb670a14d7f7b379246677826234364a

karmatosed:
There are conflicts that need resolving before we can work on this, can you take a look please @WPAddiction ?

grappler:
As there are a number of conflicts in this PR it would be best to create a new PR if need be. Please see the discussion in #1149 as to the future workings of CSS and Sass.

Issue #454, opened 2014-04-06
Menu alignment


ghost:
In section '5.2 Menus' of the style.css of _s wordpress theme the third class:

.main-navigation li {
	float:left;
bla bla bla bla other stuf
}

Because of float (left / right) you will not be albe to align your menu center because it's a list we're modifying here. To fix this, change float:left; to display:inline-block;
Now you can align your menu in .main-navigation ul (the second class in this section) using text-align (left/center/right).

Maybe usefull for you / for further updates of _s.

Jasper

YemSalat:
What's the fuss about? Why not just set the menu `ul` `font-size` to 0? and reset all `li`s You're using sass anyways,shouldn't be a problem, right?

sixhours:
https://core.trac.wordpress.org/ticket/27762 is fixed as of 3 months ago. Do we have an updated PR for `inline-block`?

obenland:
It will be reverted in 4.4.1 as it breaks back compat with menus that justify their menu items.

Anyone who wants to use inline-block for their menus will have to use their own menu walker and override the `end_el` method.

PR #894, opened 2015-12-21
Use `inline-block` rather than `float` for menu items in main navigation

sixhours:
See #454 for additional discussion.


sixhours:
Closed as per discussion in #454

PR #898, opened 2016-01-07
Adding ID to comment labels for post

helloguna1:
Each comment will reflect the post’s unique ID for better
identification.

### DIFF
--- a/template-parts/content-page.php
+++ b/template-parts/content-page.php
@@ -38,4 +38,4 @@
 			);
 		?>
 	</footer><!-- .entry-footer -->
-</article><!-- #post-## -->
+</article><!-- #post-<?php the_ID(); ?> -->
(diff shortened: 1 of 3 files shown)


karmatosed:
What is the benefit of doing this?

helloguna1:
The only difference is you get to identify which post the comment belongs to. For example, instead of `<!-- #post-## -->`, you'll get `<!-- #post-123 -->`.

It's a minor change, but I had times where I was looking at the console and had to scroll up to read which post it belonged to.

sixhours:
This makes sense to me.

helloguna1:
@karmatosed Any thoughts?

philiparthurmoore:
Looks like a great change. Travis errors need to be resolved but the change has solid reasoning behind it.

karmatosed:
Thanks for your explanation @ghost. I think lets have a refreshed PR and we can take a look.

grappler:
@karmatosed @ghost Is a deleted account. You won't get any response. We will need to close the PR and create a new one.

karmatosed:
Thanks for catching that @grappler

davidakennedy:
Closing this out as this was done in #1106.

PR #899, opened 2016-01-08
Use the H1 element for the site title when front page display is set to static

gregsullivan:
H1 tags wrapping the site title are only used when the front page is set to display WordPress's latest posts. If a site uses a static page as its front page, the conditional check will always return false.

By removing the `is_home()` conditional tag, the H1 version of the site title is used when the front page displays either a static page or the latest posts, and a P element continues to be used everywhere else.

### DIFF
--- a/header.php
+++ b/header.php
@@ -27,7 +27,7 @@
 	<header id="masthead" class="site-header" role="banner">
 		<div class="site-branding">
 			<?php
-			if ( is_front_page() && is_home() ) : ?>
+			if ( is_front_page() ) : ?>
 				<h1 class="site-title"><a href="<?php echo esc_url( home_url( '/' ) ); ?>" rel="home"><?php bloginfo( 'name' ); ?></a></h1>
 			<?php else : ?>
 				<p class="site-title"><a href="<?php echo esc_url( home_url( '/' ) ); ?>" rel="home"><?php bloginfo( 'name' ); ?></a></p>
--- a/template-parts/content-page.php
+++ b/template-parts/content-page.php
@@ -11,7 +11,12 @@
 
 <article id="post-<?php the_ID(); ?>" <?php post_class(); ?>>
 	<header class="entry-header">
-		<?php the_title( '<h1 class="entry-title">', '</h1>' ); ?>
+		<?php
+		if ( ! is_front_page() ) {
+			the_title( '<h1 class="entry-title">', '</h1>' );
+		} else {
+			the_title( '<h2 class="entry-title">', '</h2>' );
+		} ?>
 	</header><!-- .entry-header -->
 
 	<div class="entry-content">


gregsullivan:
Before submitting this pull request, I read over #776 and decided a PR was warranted since the current behaviour doesn't match the goal as described:

> Site title set to H1 on front page, P on other pages

Having looked into this a bit more, I think changes to content-page.php might also be warranted, but I want to await comments from others before updating this request. My suggestion would be to have content-page.php use logic similar to that found in content.php, with H1 and H2 being used conditionally. In the case of content-page.php, H2 would only be used when `is_front_page()` returns true; in all other situations, it would remain H1.

I think this is important for a few reasons:

* **More consistent with core** — Because core uses the site title as the HTML title on the front page regardless of whether the front page is set to display latest posts or a static page, consistently using the site title as the H1 on the front page ensures that the two are aligned.
* **Arguably better for SEO** — I've seen a number of Twenty Fifteen and Twenty Sixteen installations using static front pages with titles along the lines of "Hi there!" and "Welcome to my site." It's much less common in my experience to see the site title as the title of the page being used for the static front page since it will appear alongside the site title in the header. In my opinion, core overriding the page title in the title tag on the front page implies to users that the site title is more important in that context, which can lead to less polished page titles being placed in H1s and given more prominence than the site title. (This change would also align the site with how Yoast handles it on their site: H1 for the site title on the home page, with the page or post title taking over everywhere else, even though their front page is a static page.)
* **Ensures H1 is not overlooked in front page templates** — Because the difference between `is_home()` and `is_front_page()` is sometimes misunderstood, some developers might think they can remove the H1 from a page template built for use on the front page and that the site title will take over in that role in the same way it does when the front page displays latest posts.

I look forward to your feedback, thanks in advance!

corvannoorloos:
> It's much less common in my experience to see the site title as the title of the page being used for the static front page since it will appear alongside the site title in the header. In my opinion, core overriding the page title in the title tag on the front page implies to users that the site title is more important in that context, which can lead to less polished page titles being placed in H1s and given more prominence than the site title itself.

I definitely agree.

philiparthurmoore:
@karmatosed @iamtakashi @davidakennedy How is this being handled in Core? Thoughts?

davidakennedy:
Hey @gregsullivan, Thanks for the PR and discussion. I'll try to shed a little more light on this since I think I helped make this change. 😄 

Looking at this code by just looking at the code in the `header.php` is probably a bit misleading. You have to check out: https://github.com/Automattic/_s/blob/master/index.php#L23 as well.

These two changes were made to give _s a more traditional heading structure for better accessibility. It makes each page have one `h1`. This is better for screen reader users since they typically scan a page using a list of heading elements. It enables them to know what the most important topic on the page is and what it's about.

In HTML5, it's totally valid to have multiple `h1` headings and the browser is supposed to determine the outline based on the markup. Trouble is, [no browser has implemented it](https://www.paciellogroup.com/blog/2013/10/html5-document-outline/).

I definitely agree that there are cases where you might want your home page `<title>` to be the site title no matter what, but at least I figured it's easy to modify that with plugins. This gives you lots of control out of the box, at least.

> How is this being handled in Core? Thoughts?

@philiparthurmoore Twenty Fifteen and Twenty Sixteen follow this pattern.

Overall, I'm not in favor of merging this because if you set a static page as your home page, you have two `h1` elements – the site title and the page title. I'd rather it have a pure outline for the benefits of better structure, accessibility and seo.

All that said, it's been awhile since we made this change, so it's always good to talk about and decide if there's a better way to handle this.

gregsullivan:
Thanks for your response, @davidakennedy! I strongly agree that there shouldn't be two `h1` elements on a single page; I wanted to await comments to see if there was interest in making this change before I went ahead and updated my pull request as described in my second comment.

To give a bit of background, I've encountered this problem multiple times, always on sites using a static page as the front page while simultaneously using a custom template for that page. If the custom template doesn't include an `h1`, the site's front page will not have an `h1` at all since the one in header.php isn't used for static pages. (I originally assumed it was since I was used to seeing the site's title used as `h1` when the front page was set to show the latest posts.)

That said, I think the issues described above under "More consistent with core" and "Arguably better for SEO" are at least as compelling as ensuring `h1` isn't overlooked in custom front page templates.

My suggested amendment to this pull request is to use `h1` and `h2` conditionally in content-page.php. On a static page used as the front page, the page title would be an `h2`, and in all other circumstances it would remain an `h1`.

If you think that's something you would consider, I'd be happy to make that change.

davidakennedy:
I'd be open to exploring this change, and seeing what others think as well.

I feel like having the `h1` always be the site title on the front/home page may be a legitimate reason for a change. I'm not an expert in SEO, and was fine with having the page title be the `h1` though. 😄 

> To give a bit of background, I've encountered this problem multiple times, always on sites using a static page as the front page while simultaneously using a custom template for that page. If the custom template doesn't include an h1, the site's front page will not have an h1 at all since the one in header.php isn't used for static pages.

I will say @gregsullivan, that making a change because themers using _s have created a template that doesn't have a `h1` isn't a great reason. That seems like just poor markup to me. Are the reasons why a template wouldn't need some level of heading?

gregsullivan:
Sorry for the long delay in responding to this; I was reminded that I hadn't posted a response when I saw the merge announcement for Twenty Seventeen – congratulations, @davidakennedy! – and checked to see which approach it followed in addressing the issue described above. I was happy to see it followed the approach I've been advocating (though I don't pretend to think the above thread influenced Twenty Seventeen's approach).

> I will say @gregsullivan, that making a change because themers using _s have created a template that doesn't have a h1 isn't a great reason. That seems like just poor markup to me. Are the reasons why a template wouldn't need some level of heading?

I've encountered this a number of times on client projects since my last response. That said, a representative site that is easy to reference is [WordPress.org](https://wordpress.org/).

Imagine building a replica of WordPress.org using _s as your starting point. The homepage doesn't have a clear `h1` in its main content area (between the header and footer) because it goes straight into a description of what WordPress is; "WordPress.org" acts as the top-level heading for the page. (If you were forced to choose a different `h1`, I suppose it would be "WordPress is web software you can use to create a beautiful website, blog, or app. We like to say that WordPress is both free and priceless at the same time." That looks like a good candidate for a `p` to my eyes, though, and it is marked up that way on the existing site.)

This is the pattern I frequently encounter: Static front pages where the title of the site is designed to act as an `h1`, with a block of copy describing the company or site acting as the first copy on the page. When clients have the opportunity to enter a title for a static front page, as in themes like Twenty Seventeen, they often use something like "Welcome to my website!", which would better serve as an `h2` on the homepage of a website.

(As an example of this, the demo video for Twenty Seventeen has "Welcome to Needle & Hook Coffeehouse!" as the title on its frontpage, which I agree should be an `h2` to the `h1` of "Needle & Hook".)

As a final note, I'm not advocating for the front page template to have no heading; I think in many cases it makes sense to have an `h2` in addition to the `h1` in the `header`, and I think that should be the default for cases when the front page template is not a custom one.

In the hope you remain open to moving forward with this change, I will put together an updated pull request that makes the additional change to `content-page.php`. Thanks again for your ongoing feedback.

gregsullivan:
I've updated this pull request and squashed my commits. To reiterate the motivation for this change, these modifications will:

- Fulfill the original intent of #776 ("Site title set to H1 on front page, P on other pages")
- Match core's logic for selecting the title tag on a site's front page
- Match the behaviour of Twenty Seventeen

This benefits themes meant for widespread use by ensuring the site title is the `h1` on the front page in circumstances when users enter vague titles. It also benefits custom themes that might not use the page title when creating a custom home page template.

PR #900, opened 2016-01-13
Toggle focus class to allow submenu access on tablets

iamtakashi:
Refreshed PR #666

Tested with Chrome, Firefox, (in both iOS and Android) and iOS Safari

### DIFF
--- a/js/navigation.js
+++ b/js/navigation.js
@@ -78,4 +78,35 @@
 			self = self.parentElement;
 		}
 	}
+
+	/**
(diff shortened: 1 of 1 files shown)


sixhours:
Thank you for looking at this! :)

samikeijonen:
This would be huge! I can test it out.

sixhours:
D'oh, just realized there were no conflicts with *this* merge. Going to ditch mine and merge this one.

PR #902, opened 2016-01-14
Add support for Internet Explorer 7 and Internet Explorer 8

pmtarantino:
IE7 and IE8 do not support addEventListener, so we have to use attachEvent in this case.

### DIFF
--- a/js/navigation.js
+++ b/js/navigation.js
@@ -53,8 +53,13 @@
 
 	// Each time a menu link is focused or blurred, toggle focus.
 	for ( i = 0, len = links.length; i < len; i++ ) {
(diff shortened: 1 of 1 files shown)


davidakennedy:
Hey there @pmaiorana!

Thanks for the thoughtful pull request! I like the idea of checking for features...

That said, this isn't something we'd add as [_s supports IE 9 and up currently](https://github.com/Automattic/_s/blob/master/CONTRIBUTING.md).

Also, [Microsoft has ended support recently for IE 8, 9 and 10](https://www.microsoft.com/en-us/WindowsForBusiness/End-of-IE-support). So we'll look at when it would be good to bump up our browser support soon.

PR #905, opened 2016-01-24
Add theme support for gallery post-format.

acbilimoria:
I'm just going through the docs and thought this might be helpful.

https://developer.wordpress.org/themes/basics/theme-functions/#post-formats


karmatosed:
Adding support to any more post formats is up to the end user. _s is a starter theme and shouldn't have to or even does support everything. As a result, I am going to close this.

PR #906, opened 2016-01-25
Update the replace example in README.md

ardalann:
Many might choose multiple words as theme name, and they may not be sure where they should use a space, a hyphen or an underscore to separate words when doing the initial replacements. The example provided in the README file should cover that scenario.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@ Getting Started
 
 If you want to keep it simple, head over to http://underscores.me and generate your `_s` based theme from there. You just input the name of the theme you want to create, click the "Generate" button, and you get your ready-to-awesomize starter theme.
 
-If you want to set things up manually, download `_s` from GitHub. The first thing you want to do is copy the `_s` directory and change the name to something else (like, say, `coolest-theme-ever`), and then you'll need to do a five-step find and replace on the name in all the templates.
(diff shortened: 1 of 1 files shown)


karmatosed:
I'm not sure this is something we should add. I haven't seen someone have an issue or not know they can do this. Have you?

philiparthurmoore:
I've run into this confusion multiple times.

ardalann:
I've run into this issue many times. Whenever I wanted to use _s to develop a new theme, I had to find out where I should use a space, a hyphen or an underscore.

sixhours:
I'm fine with this change.

ianstewart:
The poor Megatherium just can't catch a break.

sixhours:
We could change it to Megatherium Is Awesome ;)

karmatosed:
I kind of like idea for prosperity and cute projectness it would be: megatherium-is-awesome

philiparthurmoore:
Let's get this in. It's a great change. Not at the computer at the moment but those Travis errors need to be looked at pre-merge.

davidakennedy:
Hey @ardalann, this is old, but worthy issue to revive. I'd like to get it merged! Can you update your change with `megatherium-is-awesome` as suggested above? Thanks!

ardalann:
Hey @davidakennedy, I just updated the example theme name.

PR #909, opened 2016-02-08
Fixed margins on Figure tags which was causing x-wide images to overflow content area.

ckschmieder:
I've set left & right margins for the 'Figure' element to 0px (originally 40px) because it was causing extra wide images (i.e. 1200px) to unintentionally overflow the content area. The 40px margin was also causing alignment issues for normal width images.

### DIFF
--- a/style.css
+++ b/style.css
@@ -151,7 +151,7 @@ svg:not(:root) {
 }
 
 figure {
-	margin: 1em 40px;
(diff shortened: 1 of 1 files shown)


ckschmieder:
Closing this request and will reissue containing a new rule for figure margins rather than changing the existing 40px from normalize

PR #931, opened 2016-03-02
Use delete_category action hook for transient flusher

jaspermdegroot:
When a category is deleted, `edit_category` isn't triggered and editing a category won't affect the count of categories with posts attached to it .

### DIFF
--- a/inc/template-tags.php
+++ b/inc/template-tags.php
@@ -117,5 +117,5 @@ function _s_category_transient_flusher() {
 	// Like, beat it. Dig?
 	delete_transient( '_s_categories' );
 }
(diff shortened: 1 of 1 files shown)


grappler:
What about when the category is edited?

jaspermdegroot:
@grappler 

Editing a category name, slug, or description won't change the number of posts attached to that category. The number of categories that have posts attached to it can only change when you edit a post or when you delete a category.

philiparthurmoore:
This makes sense. Needs testing but totally makes sense to change.

davidakennedy:
We've removed this function as a result of #1131. Closing this out. Thanks for the work here @jaspermdegroot!

PR #932, opened 2016-03-03
Merge diff between sass and css

youthkee:
When compiling Sass to CSS, there are some difference between both sources.
So, I merged to resolve conflicts.


sixhours:
I noticed this removes some of the un-styled selectors for `.main-navigation` from style.css; not sure that's a deal-breaker, but having those selectors readily available is handy if you're not using Sass.

youthkee:
It's true that there are un-styled selectors in style.css and _menus.scss both.
So I put back those selectors to style.css in places that correpond to _menus.scss.

WPAddiction:
I like the idea of reordering style.css to match the compiled SASS, but you removed some of the blank lines between selectors that should be put back before this is merged.
https://github.com/Automattic/_s/commit/9a08ca0bfb670a14d7f7b379246677826234364a

Also the explanation mark is really unnecessary in style.css and has been debated in the past.
`/*!`

youthkee:
Thank you for your information.
I Added blank line between selectors to style.css, and put back the explanation marks.

WPAddiction:
And sorry, but to clarify style.scss requires the exclamation mark to keep that comment from being stripped when compiled, but isn't needed in style.css, although I don't know that it hurts anything in style.css. Also, I'm no expert here, I've just been looking over the CSS a lot myself.

youthkee:
I checked for the exclamation mark again, and I found it's necessary in style.scss to remain the comment section when compiling.
So, I put back this one.
e9c7d73be804365e44934f2b01b70e0110bcef13

joshmcrty:
I saw this PR and was wondering, has anyone looked into compiling the sass on underscores.me instead and removing the CSS file from this repo? The intent is that developers generally download the theme from underscores.me and not directly from this repo, correct? It already handles find/replace for theme names, couldn't it handle creating the CSS file if you don't check the "sassify" box on underscores.me? That would eliminate some of the out-of-sync issues this repo has had in the past with CSS/Sass pull requests.

grappler:
@joshmcrty #1159 #1149 is looking at automating the CSS generation which may fix the out of sync issue without needing to generate the CSS on theme generation.

davidakennedy:
Is this still an issue? When I ran a test compile yesterday, the only differences I saw were stylistic ones because of the Sass `expanded` output.

youthkee:
I tried to compile with latest codes, and there seems to be one difference at a comment line break yet. Is it better to fixed that?

I compiled with node-sass and used options below.

```
$ node-sass --output-style expanded --indent-type tab --indent-width 1 --precision 10 sass/style.scss style.css
```

Ismail-elkorchi:
Closing as this issue is addressed in https://github.com/Automattic/_s/pull/1391.

PR #935, opened 2016-03-07
Remove superfluous check from comment template

jaspermdegroot:
If the post type doesn't support comments `get_comments_number()` will already return 0

### DIFF
--- a/comments.php
+++ b/comments.php
@@ -79,15 +79,13 @@ if ( post_password_required() ) {
 		<?php
 		endif; // Check for comment navigation.
 
+		// If comments are closed and there are comments, let's leave a little note, shall we?
+		if ( ! comments_open() ) : ?>
+			<p class="no-comments"><?php esc_html_e( 'Comments are closed.', '_s' ); ?></p>
+		<?php
+		endif;
+
 	endif; // Check for have_comments().
-
-
-	// If comments are closed and there are comments, let's leave a little note, shall we?
-	if ( ! comments_open() && get_comments_number() && post_type_supports( get_post_type(), 'comments' ) ) : ?>
-
-		<p class="no-comments"><?php esc_html_e( 'Comments are closed.', '_s' ); ?></p>
-	<?php
-	endif;
 
 	comment_form();
 	?>


jrfnl:
This check is not superfluous. You may have a post type which does not support comments at all, in that case, echo-ing out 'no comments' just doesn't make sense.

jaspermdegroot:
@jrfnl 

The message is not "no comments", but "Comments are closed". The message is meant for when there are comments being displayed, but commenting is no longer possible. If a post type supported comments in the past, no longer supports them, but there are comments already, those comments will be displayed. So the note still makes sense.

In other words, the only thing that matters for displaying the "Comments are closed" or not is whether there are comments being displayed. `get_comments_number()` takes care of that.

jrfnl:
I stand corrected.

freddy0077:
i dont care this repo is shit

philiparthurmoore:
> If the post type doesn't support comments get_comments_number() will already return 0

@jaspermdegroot Can you validate this with a link to Core code or make this much easier to figure out by adding some context? It's hard to consider pull requests with such terse explanations. Thanks so much.

philiparthurmoore:
Also, what happens if a plugin enables comments on a post type that by default doesn't support them?

jaspermdegroot:
@philiparthurmoore 

Thanks for your feedback. My initial comment should indeed have contained a better explanation. Sorry about that.

> If the post type doesn't support comments get_comments_number() will already return 0

I realised that this is not always true after @jrfnl her comment about "a post type which may have supported comments in the past, but no longer does so".

Anyway, it's not relevant if a post type still supports comments or only supported comments in the past and if this support was enabled by default or by a plugin. The only thing that matters is if there are comments being displayed or not. In that case we want to display the "Comments are closed" message if commenting is no longer possible.

That being said, the code can be further simplified by putting it inside the `have_comments()` check. I will push a commit for that.

comments.php
```diff
@@ -69,15 +69,14 @@
...
-	endif; // Check for have_comments().
-
+		// If comments are closed and there are comments, let's leave a little note, shall we?
+		if ( ! comments_open() ) : ?>
 
-	// If comments are closed and there are comments, let's leave a little note, shall we?
-	if ( ! comments_open() && get_comments_number() && post_type_supports( get_post_type(), 'comments' ) ) : ?>
+			<p class="no-comments"><?php esc_html_e( 'Comments are closed.', '_s' ); ?></p>
+		<?php
+		endif; // Check for comments closed.
```
grappler: The closing comment is normally just added if there is a long if statement. In this case I don't think it is needed and I am not sure if it grammatically correct.

comments.php
```diff
@@ -69,15 +69,14 @@
 		<?php
 		endif; // Check for comment navigation.
 
-	endif; // Check for have_comments().
-
+		// If comments are closed and there are comments, let's leave a little note, shall we?
+		if ( ! comments_open() ) : ?>
```
grappler: Can we remove this extra space too?

jaspermdegroot:
@grappler - Thanks for reviewing this PR. I pushed another commit to remove the new line and the closing comment.

grappler:
@jaspermdegroot Great! Do you think you could squash all of the commits into one? This article explains how to do it. https://www.devroom.io/2011/07/05/git-squash-your-latests-commits-into-one/ You will need force push the changes to GitHub. You can do it with this command `git push --force`.

jaspermdegroot:
@grappler - Sure, I will do an interactive rebase and force push.

jaspermdegroot:
@grappler - I will do another rebase to update my email address to the one I currently use in my Git config

PR #938, opened 2016-03-22
Add widget description

emiluzelac:
_From the theme review standpoint (both marketplace and directory)_

I keep seeing authors leave description as-is (empty) and believe that it wouldn't hurt to add something in by default. This is a general description taken from https://github.com/WordPress/twentysixteen/blob/master/functions.php#L150

At the same time it shows good practice how to escape the content and translate the string too.


karmatosed:
I'd almost not say sidebar incase it's not used for an actual sidebar. Want to refresh the patch leaving that out?

emiluzelac:
Like that?

karmatosed:
Yep that works well, thanks!

PR #940, opened 2016-04-05
Add template tag for entry titles for better consistency

johncionci:
Create a template tag to render site titles for cleaner theme code.


philiparthurmoore:
In principle this makes sense but wouldn't `is_singular` be better here?

grappler:
I don't think this is needed. The title in `content-page.php` will always be without a link and `content-search.php`  will need a link. The if else in `content.php` is sufficient.

alaczek:
Agree with @grappler here - it feels like a function in this case makes things more complicated then necessary.

PR #942, opened 2016-04-12
abbr needs text-decoration removed for FF. pre & abbr were styled in 2 places

andrewklimek:
abbr elements had double lines in Firefox because FF applies text-decoration: dotted and are the only browser supporting it.

### DIFF
--- a/style.css
+++ b/style.css
@@ -99,10 +99,6 @@ a:hover {
 	outline: 0;
 }
 
-abbr[title] {
-	border-bottom: 1px dotted;
-}
-
 b,
 strong {
 	font-weight: bold;
@@ -159,13 +155,8 @@ hr {
 	height: 0;
 }
 
-pre {
-	overflow: auto;
-}
-
 code,
 kbd,
-pre,
 samp {
 	font-family: monospace, monospace;
 	font-size: 1em;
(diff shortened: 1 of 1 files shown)


corvannoorloos:
Personally I'd like to keep normalize.css untouched. (perhaps minus certain browser fixes if eventually automated)

andrewklimek:
Ok, I can see that.  Maybe then remove the duplicate attributes from further down?

Either way the text-decoration: none for firefox would be a great reset addition.

corvannoorloos:
I believe the text decoration is actually there for a reason. It marks the existence of a title attribute. (similar as a link, but without the source)

The `abbr` tag needs to be updated though. I believe recent changes have been made [here](https://github.com/necolas/normalize.css/pull/468) where they went for `underline` instead of `border-bottom`.
A few commits after they've also normalized the `underline` to behave similar in different browsers.

An update of _s' nomalize.css might be the best solution here.

philiparthurmoore:
Agreed that it'd be great if normalize were left as-is. It'll make it a lot easier to drop in updates when they happen.

andrewklimek:
Ok makes sense in that regard.

PR #943, opened 2016-04-15
calculate gallery image size

marcobiedermann:
Calculate gallery image size instead of hardcoding the values is more maintainable and understandable.

Btw. which tool do you use to compile the Scss files?

### DIFF
--- a/sass/media/_galleries.scss
+++ b/sass/media/_galleries.scss
@@ -6,38 +6,38 @@
 	display: inline-block;
 	text-align: center;
 	vertical-align: top;
-	width: 100%;
+	width: (100% / 1);
 
 	.gallery-columns-2 & {
-		max-width: 50%;
+		max-width: (100% / 2);
 	}
 
 	.gallery-columns-3 & {
-		max-width: 33.33%;
+		max-width: (100% / 3);
 	}
 
 	.gallery-columns-4 & {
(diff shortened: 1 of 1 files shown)


sixhours:
This works for me in testing! Good idea.

I'm not sure why we're dividing 100% by 1 in the parent `width` attribute; would there ever be a situation where we'd want to calculate a different value there? Doesn't hurt anything, but seems unnecessary.

We also need to compile the Sass and update style.css to match before we merge this.

marcobiedermann:
@sixhours 
> I'm not sure why we're dividing 100% by 1 in the parent width attribute;

You are absolutely right, dividing by 1 is irrelevant to the calculation but I used to intentionally to create a pattern, which is a bit easier to follow (dividing by 1, 2, 3, 4 …).

> We also need to compile the Sass and update style.css to match before we merge this.

Yes sure, but I do not know which compiler is used for this project.

sixhours:
> Yes sure, but I do not know which compiler is used for this project.

Good question....I think we've been updating style.css manually for changes like this, but someone else may be able to confirm.

philiparthurmoore:
Let's close this and do it the #986 way. Sound good?

marcobiedermann:
@philiparthurmoore I agree!

PR #947, opened 2016-04-25
Custom header re-wording

emiluzelac:
A custom header is no longer commented: https://github.com/Automattic/_s/blob/master/functions.php#L132


emiluzelac:
Thank you 👍

PR #949, opened 2016-04-28
Only show page entry footer if edit link is available

phoenixenero:
Adds a condition which checks the output of `get_edit_post_link()` before displaying the entry
footer.

### DIFF
--- a/template-parts/content-page.php
+++ b/template-parts/content-page.php
@@ -25,6 +25,7 @@
 		?>
 	</div><!-- .entry-content -->
 
+	<?php if ( get_edit_post_link() ) : ?>
 	<footer class="entry-footer">
 		<?php
 			edit_post_link(
@@ -38,4 +39,5 @@
 			);
 		?>
 	</footer><!-- .entry-footer -->
+	<?php endif; ?>
 </article><!-- #post-## -->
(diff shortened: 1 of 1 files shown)


sixhours:
This works for me in testing. I suggest we merge it. Cleaner markup for logged-out users. 👍

jrfnl:
Hmm.. in effect you're running the `get_edit_post_link()` function twice now which may not be the most efficient.

Also, the `_s_entry_footer()` function contains the same function call with the exact same parameters, although without the superfluous markup issue. Candidate for it's own template tag function ?

phoenixenero:
Honestly I don't care about efficiency considering the template appears only on pages – plus most people (I hope) usually have their own caching solution.

As for adding a template tag for page footers, I think that should be left up to the theme developer.

davidakennedy:
I'd be in favor of merging this as well. Tested, and works well. @phoenixenero Can you update the PR to indent the HTML `<footer class="entry-footer">` inside the `<?php if ( get_edit_post_link() ) : ?>` statement? That way, everything is indented properly. Thanks!

phoenixenero:
Just indented the `.entry-footer` block; you can merge now.

PR #956, opened 2016-05-16
Narrowing the scope of the transient flusher to post types that have category as a taxonomy

emhr:
Currently in s_category_transient_flusher, the transient is flushed whenever any post type is saved
(pages for example). This narrows the focus to posts only using the
save_post_{post_type} hook which has been around since WP 3.7

### DIFF
--- a/inc/template-tags.php
+++ b/inc/template-tags.php
@@ -119,4 +119,18 @@ function _s_category_transient_flusher() {
 	delete_transient( '_s_categories' );
 }
 add_action( 'edit_category', '_s_category_transient_flusher' );
-add_action( 'save_post_post', '_s_category_transient_flusher' );
+
+/**
+ * Queue the add_action for save_post_{post_type} only if the post type has the 'category' taxonomy
+ */
+function _s_queue_transient_flusher() {
+	$post_types = get_post_types( array( 'public' => true, '_builtin' => true ), 'names' );
+	foreach ( $post_types as $post_type ) {
+		$taxonomies = get_object_taxonomies( $post_type );
+		if ( in_array( 'category', $taxonomies ) ) {
+			add_action( 'save_post_' . $post_type, '_s_category_transient_flusher' );
+		}
+	}	
+}
+// Hooking late on init to let plugins register custom posts types first
+add_action( 'init', '_s_queue_transient_flusher', 9999 );
--- a/inc/template-tags.php
+++ b/inc/template-tags.php
@@ -119,4 +119,4 @@ function _s_category_transient_flusher() {
 	delete_transient( '_s_categories' );
 }
 add_action( 'edit_category', '_s_category_transient_flusher' );
-add_action( 'save_post_post',     '_s_category_transient_flusher' );
+add_action( 'save_post_post', '_s_category_transient_flusher' );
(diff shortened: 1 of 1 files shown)


phoenixenero:
Minor nitpick but the non-alignment of the 2nd arguments is kinda... distracting? You could remove the extra spaces between `'save_post_post',     '_s_category_transient_flusher'` to bring it down to one space.

emhr:
Right on. The spacing is corrected now.

emhr:
Now Checking registered post types and only adding the save_post_{post_type} action if the post type has the category taxonomy.

inc/template-tags.php
```diff
@@ -119,4 +119,18 @@ function _s_category_transient_flusher() {
...
+
+/**
+ * Queue the add_action for save_post_{post_type} only if the post type has the 'category' taxonomy
+ */
+function _s_queue_transient_flusher() {
+	$post_types = get_post_types( array( 'public' => true, '_builtin' => true ), 'names' );
+	foreach ( $post_types as $post_type ) {
+		$taxonomies = get_object_taxonomies( $post_type );
+		if ( in_array( 'category', $taxonomies ) ) {
+			add_action( 'save_post_' . $post_type, '_s_category_transient_flusher' );
```
jrfnl: You should add a `return;` after this line as once the action has been added, there is no need to continue walking through the array.

phoenixenero: ...What about the other post types? Returning early would disregard the other post types with the category taxonomy.

jrfnl: Oh darn.. you're right of course. I completely missed the variable part of the hook. Blaming my dyslexia and crawling away to a corner again ;-)

emhr:
@grappler you had some reservations on the earlier approach mentioned in Issue #959 what's your opinion of this now?

philiparthurmoore:
@grappler ^

grappler:
Sorry this went under. Have you tested if `$post_types = get_post_types( array( 'public' => true, '_builtin' => true ), 'names' );` also returns a CPT. The reason I ask is that I think in this case the function will only return Post Types that are public **and** builtin and not public **or** builtin.

We could fix this by defining the operater as or `$post_types = get_post_types( array( 'public' => true, '_builtin' => true ), 'names', 'or' );`

I think it would be also more performant if we kept the action as `save_post` and just called `_s_category_transient_flusher ` within `_s_queue_transient_flusher()`. The current method will require running the whole checks on every page load and not just when you are saving a post.

davidakennedy:
@emhr Interested in continuing the work here based on the above feedback? It would be nice to work it in with the changes done to the function in Twenty Seventeen. See: https://core.trac.wordpress.org/browser/trunk/src/wp-content/themes/twentyseventeen/inc/template-tags.php#L154

emhr:
@davidakennedy 
sure thing

davidakennedy:
We decided to remove the `s_category_transient_flusher` function because of #1131.

We added a new feature from Jetpack, Content Options. The category functions weren't needed since we now show categories whether a site has one or more than one. Also, many customers on WordPress.com found it confusing that themes only showed categories if a site had more than one.

Thanks for the work here @emhr! Closing this out – hope to see your work in another pull request!

PR #962, opened 2016-05-26
Fix menu touch taps

sixhours:
Add JS function to handle toggling `.focus` class on parent menu items so child menus can be accessed on tablets. Props @iamtakashi


sixhours:
See #666 and #900 for reference.

sixhours:
Duplicate of #900, closing!

PR #964, opened 2016-05-27
Comment the large menu & small menu styling

bahiirwa:
Added comment the lines to help the builder make meaningful menu. 
Also add code important for the small menu styling without sending the user searching over the globe.

### DIFF
--- a/style.css
+++ b/style.css
@@ -532,6 +532,9 @@ a:active {
 /*--------------------------------------------------------------
 ## Menus
 --------------------------------------------------------------*/
+/*
+ * .main-navigation Target the menu
+*/
 .main-navigation {
 	clear: both;
 	display: block;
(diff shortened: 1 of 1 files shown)


bahiirwa:
I think this is the design element missing in the artillery.
Added comment the lines to help the builder make meaningful menu. 
Also add code important for the small menu styling without sending the user searching over the globe.

phoenixenero:
I think comments are superfluous. The selectors describe themselves already.

I prefer (and so does the [WP CSS coding standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/css)) each selector be kept to its own line:

```
/* yes */
.selector-a,
.selector-b,
.selector-c {
    /* ... */
}

/* no */
.selector-a, .selector-b, .selector-c {
    /* ... */
}
```

I have mixed feelings about duplicating the selectors in the small menu media query. While it is kinda useful, the developer can just copy and paste the large menu styles into the small menu query.

bahiirwa:
@phoenixenero Adjustments can be made to the Coding standard & small menu Duplication. My initial thoughts on commenting were to make it simple for the new user to identify the target since they might have issues with the WP structure while implementing complex menus.

sixhours:
I personally don't feel the comments are necessary, but if we were to add something like this, it should be added to the corresponding Sass as well.

karmatosed:
I'd vote against the comments also in this case. Thanks for suggesting though.

PR #965, opened 2016-05-28
Only load pingback URL if it's a singlar page

HardeepAsrani:
Only load pingback URL if it's a singular page and pings are open. Twenty Sixteen is doing the same. Similar to #853 but a better way to do it.


jrfnl:
See the discussion in #853. The original PR was the same as yours. It was adjusted based on the feedback received.

HardeepAsrani:
@jrfnl Sorry didn't notice that at first. I will close it. :)

PR #970, opened 2016-06-14
Escape header attributes.

ThierryA:


emiluzelac:
It is not necessary to escape them: https://core.trac.wordpress.org/browser/tags/4.5/src/wp-includes/general-template.php#L645 they're already filtered.

ThierryA:
@emiluzelac if I may say, we can't predict which part of a website will be hacked and what will the hacker have access to. For example a website could have a leek through `bloginfo` filter **only** (due to a plugin for example) allowing someone who would have gained access to that part of their website and modify what is returned. It would have the same effect as doing something like (just example of the potential output):

```
add_filter( 'bloginfo', function( $value, $show ) {
	if ( 'charset' === $show ) {
   		return '"><script>alert("You have been hacked.");</script>"<';
   	}

   	return $value;
}, 10, 2 );
```

This would not be vulnerable if the front end was safely escaped using `<?php esc_attr_e( get_bloginfo( 'charset' ) ); ?>`.

My point is that it is very difficult to predict exactly what part of the system hackers will gain access to and making the front as safe as possible can prevent such things.

emiluzelac:
Cool, that filter link is even better than mine 👍

karmatosed:
I think what is being said is to discard this, if not then let me know.

PR #980, opened 2016-06-29
Navigation - menu ancestor fix

roarytubbs:
Renamed
.current_page_ancestor > a
to
.current-page-ancestor > a,

### DIFF
--- a/sass/navigation/_menus.scss
+++ b/sass/navigation/_menus.scss
@@ -63,7 +63,8 @@
 
 	.current_page_item > a,
 	.current-menu-item > a,
-	.current_page_ancestor > a {
+	.current-page-ancestor > a,
+	.current-menu-ancestor > a {
 	}
 }
 


karmatosed:
Does this fix a particular ticket or issue? Would be great to see context here, thanks.

philiparthurmoore:
@roarytubbs ^

roarytubbs:
This was just a simple css fix. The current class was not named properly.

gatespace:
Looking at the reference seems to correct the `current_page_ancestor`.
see:
https://developer.wordpress.org/reference/functions/wp_nav_menu/#backward-compatibility-with-wp_page_menu

Does this change is necessary?
What is WordPress version?

bahiirwa:
Check #998 there is a similar issue

davidakennedy:
Thanks for the contribution @roarytubbs! As far as I can tell, the current classes should cover all the needed situations. This was brought up in another issue as well. See: https://github.com/Automattic/_s/issues/1073#issuecomment-272638073

I'm going to close this out, but if I'm missing something, feel free to comment and I'll take a look.

PR #992, opened 2016-07-25
Create custom background.php

bahiirwa:
Add functionality to utilize the wordpress custom background feature.


grappler:
`_s` already supports the custom background https://github.com/Automattic/_s/blob/master/functions.php#L62

bahiirwa:
I think I had a bug in mine.. Thanks

PR #993, opened 2016-07-29
Responsive method for _s

joviczarko:
I assume it is a bit cocky to propose this, but during education of people on several meetups and workshops I noticed they are hard adopters of responsive design from the start. Maybe we need to push them a bit further and to set a base for responsive _s.

Also, I am sorry if I am stupid and this is not the way it should be done. I have tried this method in my work and with attendees of the wp gatherings, and it seems they adopt it well.

Also, I am aware that _s is just a pure skeleton for wp theme, and this is maybe an overhead, but it maybe help someone.

Short description of the method:

Added responsive folder to the sass folder. Main _responsive.scss file is called from style.scss. Responsive file contains media queries which calls the individual files for every screen size. Break points are easily changeable via variables in _structure.scss file in the variables folder.

Advantages:
- Every theme will be already prepared for the responsive
- No worries about braces {} of the media queries
- Easy changeable break points via variables

**P.S. This change has no huge effect on the CSS file or we can maybe include the empty media queries.**

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,0 @@
-style.css.map
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,1 @@
+style.css.map
--- a/sass/forms/_buttons.scss
+++ b/sass/forms/_buttons.scss
@@ -6,17 +6,21 @@ input[type="submit"] {
 	border-color: $color__border-button;
 	border-radius: 3px;
 	background: $color__background-button;
+	box-shadow: inset 0 1px 0 rgba(255, 255, 255, 0.5), inset 0 15px 17px rgba(255, 255, 255, 0.5), inset 0 -5px 12px rgba(0, 0, 0, 0.05);
 	color: rgba(0, 0, 0, .8);
 	@include font-size(0.75);
 	line-height: 1;
 	padding: .6em 1em .4em;
+	text-shadow: 0 1px 0 rgba(255, 255, 255, 0.8);
 
 	&:hover {
 		border-color: $color__border-button-hover;
+		box-shadow: inset 0 1px 0 rgba(255, 255, 255, 0.8), inset 0 15px 17px rgba(255, 255, 255, 0.8), inset 0 -5px 12px rgba(0, 0, 0, 0.02);
 	}
 
 	&:active,
 	&:focus {
 		border-color: $color__border-button-focus;
+		box-shadow: inset 0 -1px 0 rgba(255, 255, 255, 0.5), inset 0 2px 5px rgba(0, 0, 0, 0.15);
 	}
 }
--- a/sass/forms/_buttons.scss
+++ b/sass/forms/_buttons.scss
@@ -6,21 +6,17 @@ input[type="submit"] {
 	border-color: $color__border-button;
 	border-radius: 3px;
 	background: $color__background-button;
-	box-shadow: inset 0 1px 0 rgba(255, 255, 255, 0.5), inset 0 15px 17px rgba(255, 255, 255, 0.5), inset 0 -5px 12px rgba(0, 0, 0, 0.05);
 	color: rgba(0, 0, 0, .8);
 	@include font-size(0.75);
 	line-height: 1;
 	padding: .6em 1em .4em;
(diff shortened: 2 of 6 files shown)


bahiirwa:
This is good idea worth thinking about. Although some developers like myself use frameworks like bootstrap or purecss to include responsiveness. How do you reconcile preference of frameworks and the redundant css? Your PS above is good.

joviczarko:
Hey @bahiirwa, 

Well, I also like using frameworks (especially bootstrap), and this method is just going perfect with bootstrap. It is NOT inteded to replace the frameworks, but to be a good addition to them.

I firstly call bootstrap in my header (wp_head), then my compiled css. I do not touch bootstrap, and I am adding my css to the my sass (compiled in style.css). I am starting with the mobile version, and fill all SASS files with my default values for mobile devices. Then, I go to develop for tablet (small) and I am entering modifications to the small.scss

What are advantages?
Your settings are saved in your css not in framework's. So you can change your framework, and just change breaking media query points and your css remains safe.

Bottom line. This has great advantages in using it with framework. Actually, framework has more to the with your markup, then css. Also, if you do not want to use the framework, you can just continue using this responsive method to develop responsive design without frameworks. It is your choice.

josephfusco:
Interesting concept @joviczarko . I like it 👍  
With how how differently CSS architecture is for everyone, and while this this is a great way to handle things, it is still an opinionated and might not be what other developers want to go with.

Maybe we can put some different recommended ways of handling things like this in the project README or wiki?

slushman:
I agree with @josephfusco. Since _s is a starter theme, these kinds of decisions should be left up to each developer and/or project. Personally, I mix my media queries in with everything else so I can see what happens to each element while I'm editing it.

joviczarko:
Hey @josephfusco,

Thanks for your kindly comments. I agree with you. For experienced developers maybe it is not suitable, since they maybe have their own ways of setting things up. Anyway, compiling empty media queries with SASS do not produces any output to the css, so it is not very important if they are using this method or not. It will not affect any development process if you just skip settings of it.

This was meant to target beginners who are still not aware of responsiveness and how to do it so they skip it always. But also to make a standardized practice for more experienced developers.

I also agree with @slushman for making queries in every scss file so he can track  changes. But... You can continue to make your media queries for elements using breaking point variables so you can control al your media queries from one plase, and you can put general elements definition in files I proposed.

So, we need to decide whether _s is oriented towards experienced developers or to support beginners without any affect to the experienced developers.

I now see this is an interesting topic which needs more participants, so we can finally solve the problem of responsiveness with beginners.

Anyhow I made a small repository as an add on for the _s 
https://github.com/joviczarko/Responsive-SASS-add-on-for-_s-theme

Also if you think I can help you with anything you had in mind, just write. I will be happy to contribute.

damiele:
Hi everyone,

I'm the unexperienced designer/developer, this my feedback. Please feel
free to ask for more.

For my next project (and all WPs coming next) I chose _s. This choice is my
most significant leap forward in understanding Wordpress (and developing
serious projects). I paired it with Skeleton CSS framework.

I didn't find the experience challenging, and honestly I think _s provide a
very well managed style.css to do your things without being forced to
massive rewritings.

I think as well that things may be improved, from an education point of
view. I slightly regret Skeleton, for example. Maybe a short guide, or a
cheatsheet will improve things, but is it worth it? To me a shortlist of
framework to pair would have been of help time ago, but I totally
understand this can't be _s Team responsibility. As well I think I prefer
Team to be involved and focused on developing what's next and not educating
a unserbase that can learn in thousands other places. But please understand
this is my personal opinion.

This being my first message I also want to thank everyone for your efforts,
it really did change my, well, everything related to design/developing.

Have a great summer you all!

*Daniele Di Stanio*
fabbricacinema.com <http://www.fabbricacinema.com>, Founder / CdA
alberodelcolore.com <http://www.alberodelcolore.com>, Founder
colorduels.com <http://www.colorduels.com>, Founder

karmatosed:
I don't feel we need to do this at this time. The idea is the theme simply is a starting point. There are also a lot of debates about approaches like this.

For now, lets keep it as it is. If someone does want to create a fork or do things this way as with anything _s that's something anyone can do.

PR #999, opened 2016-08-04
Inconsistent Markup on search page

ragzor:
all files are using a div for #primary except for the search page.


karmatosed:
There is no reason we can't have section here and actually according to the spec that actually makes sense: http://html5doctor.com/the-section-element/

PR #1001, opened 2016-08-07
Make the `content.php` more flexible

Clorith:
The fallback `content.php` file currently only works out of the box for non-hierarchical custom post types (similar to posts), as there's a `content-page.php` shipped with _s.

Since CPTs can be hierarchical as well, it makes sense to check for `is_singular()` here to avoid outputting linked `<h2>` tags on non-archive pages.


davidakennedy:
This makes a lot of sense @Clorith! Thanks for the pull request and welcome to _s. 😄

PR #1007, opened 2016-08-17
Fix typo

B-07:
line no 11, it says `_s` -based theme. I think the `-` before `based` is not necessary there


grappler:
Does the search and replace on underscores.me still work with this change? In this case the `_s` should not be replaced by the new theme name.

B-07:
Not sure what you mean.

grappler:
When generating a theme on underscores.me the scripts looks for specific cases of `_s` to replace with the theme name that you have defined in the form. Maybe the dash was a way to prevent this instance of `_s` to be replaced as if it was replaced the sentence would not make sense.

B-07:
This is in CONTRIBUTING.md file, so I guess this won't replaced by any script while generating a new theme :)

PR #1012, opened 2016-08-25
Fix translatability of comments title.

jrfnl:
As per the guidelines in https://codex.wordpress.org/I18n_for_WordPress_Developers#Plurals

> Note that some languages use the singular form for other numbers (e.g. 21, 31 and so on, much like '21st', '31st' in English). If you would like to special case the singular, check for it specifically

This fixes an error notice which WPCS will start throwing once the - soon to be released - 0.10.0 version has become stable.

### DIFF
--- a/comments.php
+++ b/comments.php
@@ -30,11 +30,13 @@ if ( post_password_required() ) {
 				$comment_count = get_comments_number();
 				if ( 1 === $comment_count ) {
 					printf(
+						/* translators: 1: title. */
 						esc_html_e( 'One thought on &ldquo;%1$s&rdquo;', '_s' ),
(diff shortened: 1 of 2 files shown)


jrfnl:
All: the travis build of the `master` branch is currently failing as it is now tested against the 0.10.0 version of WPCS. This PR will fix that.

karmatosed:
Can we get a refresh on this patch as it has conflicts?

jrfnl:
Refreshed (.pot file)!
N.B. I added one extra commit to add the missing translators comments to the new strings.

balcsida:
Merge maybe?

davidakennedy:
Thanks for the work here @jrfnl!

PR #1014, opened 2016-08-29
Enqueue Font Awesome for use

bahiirwa:
This does not add to the core functionality of _S theme but I think it is becoming a "necessary evil". Since DashIcons is not increasing the number of growing social media icons and several others, developers feel like they need to have the freedom to express their work with more icons. Enqueueing Font Awesome for use in WP Admin and Template would help a lot to solve this issue.

Line 124-131 in the template-tags.php


bahiirwa:
Please check this commit https://github.com/Automattic/_s/pull/1014/commits/b53bae5e55932ee1dfe1201b0ae0d2ceb6333114
 and we can trow ideas about this.

grappler:
This would be a no from me as I don't think every theme needs font icons. I also think that font icons are going out of fashion in favour of SVG. https://sarasoueidan.com/blog/icon-fonts-to-svg/

I have seen theme authors create their own icons to match the design of the theme.

The theme review team does not allow themes to enqueue assets from a CDN so that would be something that would need to be changed if it were to be merged.

bahiirwa:
Good insights from the link that you shared and https://css-tricks.com/icon-fonts-vs-svg/ However I think font awesome is progressive and will soon tackle the issues rising. SVGs still require for extra development time to refine for yet font icon are ready for use straight out of the box.

I think this sparks off somethings to think about. I wonder what the others think

David-Brown:
I use FontAwesome a lot, but it's not the right solution for every project and, as noted previously, not every project uses icon fonts at all. I'm not sure adding the weight of an extra library would be a good choice when some developers might want something different, or want to manage it differently (i.e.: Bower or another package manager).

bahiirwa:
Can't we compare the extra resources like custom background, sidebar CSS in layouts and header in the build also baggage. In most cases, I don't use them. But they can be commented out just like implementation of font awesome. It's making a rich resource available for use.

josephfusco:
I also believe this should not be included into _s. While Font Awesome is useful, fonts & icon fonts should be chosen on a developer level per project.

David-Brown:
@bahiirwa I think I would consider those as options within the theme, whereas an icon font is a dependency best managed according to the developer's preference.

joshmcrty:
My opinion is that Font Awesome should not be included in _s. This seems to be too heavy-handed in terms of making a design decision for the theme author. I agree with the reasons others have commented as well. It's not a core feature like custom backgrounds, etc., and it's another external dependency. Personally I've used Font Awesome for several projects, but it's not a good fit for all or even most projects in my experience.

PR #1017, opened 2016-09-09
Allow for content-{post_type}.php template files in single.php

mrwweb:
I often want to modify the contents of `content.php` for a specific post _type_. In fact, for me, this is more common than wanting to modify `content.php` by post _format_. Currently, that means having to create a new `single-{post_type}.php` file, often for the sole purpose of changing the `get_template_part()` call.

This change allows for the use of `template-parts/content-{post_type}.php` (e.g. `content-jetpack-portfolio.php`) without having to create separate `single-{post_type}.php` files. What previously required two new files and much more duplicated code, now only takes one. This PR maintains the existing `content-{post_format}.php` file ability while adding the `post_type` extensions.

### DIFF
--- a/single.php
+++ b/single.php
@@ -15,7 +15,7 @@ get_header(); ?>
 		<?php
 		while ( have_posts() ) : the_post();
 
-			get_template_part( 'template-parts/content', get_post_format() );
+			get_template_part( 'template-parts/content', get_post_type() );
 
 			the_post_navigation();
 


mrwweb:
Hmmmm. Struggling to understand the failed check. Is Travis-CI getting tripped up by the ternary syntax?

sheabunge:
Perhaps you should be using `===` for comparison?

slushman:
Great idea! What about putting the result of get_post_type() into a variable, then use that in the rest? It would avoid calling the same function twice.

`$post_type` = get_post_type();
get_template_part( 'template-parts/content', 'post' == $post_type ? get_post_format() : $post_type );'

mrwweb:
Those both seem worthwhile. Will do.

grappler:
The test is failing as this PR has not been merged yet https://github.com/Automattic/_s/pull/1012

mrwweb:
Broke this out for readability. Hopefully that would make people more likely to immediately understand it and then use it!

`$post_type = get_post_type();
$template_slug = 'post' === $post_type ? get_post_format() : $post_type;
get_template_part( 'template-parts/content', $template_slug );`

jrfnl:
Why not get rid of the ternary altogether and use a straight if/else construct ?

mrwweb:
As requested, I dropped the ternary logic and switched the check to use `get_post_format()` since that returns false when no post format is set:

```php
if( $post_format = get_post_format() ) {
	get_template_part( 'template-parts/content', $post_format );
} else {
	get_template_part( 'template-parts/content', get_post_type() );
}	
```

jrfnl:
Just a personal opinion, but I'm not a fan of assignments within conditions as they obfuscate the logic and often lead to hard to debug bugs.

mrwweb:
Switched back to a post_type check since that's a clearer description of the purpose of this setup. Dropped `$post_type` variable which I think is negligible in terms of performance. Leaves us with probably the WordPressiest implementation yet! What do we think?

```php
if ( 'post' === get_post_type() ) {
	get_template_part( 'template-parts/content', get_post_format() );
} else {
	get_template_part( 'template-parts/content', get_post_type() );
}
```

joshmcrty:
@mrwweb this is a great idea, and I think the latest version is very WordPressy indeed

truongwp:
@mrwweb Great idea! But what happen if custom post type name is as same as post format name?

mrwweb:
@truongwp I haven't run into that, largely because the more generic my post type name (e.g. video), the more important it seems that it be prefixed.

For instance, not that it's a collision with Post Formats, but this setup lets me use custom single post content templates with Jetpack's Portfolio with the file `content-jetpack-portfolio.php`. I feel like it's fair to assume developers will prefix like that. If not, you can always prefix or suffix it however you want (`content-type-*.php` / `content-format-*.php`).

Most popular plugins I see prefix their post types, and that seems especially important for plugins.

It's a good point, but I'd vote we just let theme developers handle it. Am I being too optimistic?

truongwp:
@mrwweb I agree. And this is a starter theme, developers can feel free to change it, so this is not a big problem.

alaczek:
@mrwweb As per discussion in #1148 could you update this PR and remove the `get_post_format()`?

mrwweb:
Hadn't seen that ticket, @alaczek. I'm all for it. Will get an update together soon.

mrwweb:
Updated PR so it's just this:

`get_template_part( 'template-parts/content', get_post_type() );`

Not sure why Travis CI is failing.

single.php
```diff
@@ -15,8 +15,8 @@
 		<?php
 		while ( have_posts() ) : the_post();
 
-			get_template_part( 'template-parts/content', get_post_format() );
-
+			get_template_part( 'template-parts/content', get_post_type() );
+
```
grappler: @mrwweb This is why. You need to delete the extra tabs

grappler:
@mrwweb Do you think you could squash all of the commits into one? This article explains how to do it. https://www.devroom.io/2011/07/05/git-squash-your-latests-commits-into-one/ You will need force push the changes to GitHub. You can do it with this command git push --force. If you would like to use another editor other than vim you could run git config --global core.editor "nano"

mrwweb:
Working on figuring out the squash thing. In the mean time, seems like it's [possible through GitHub when committing](https://help.github.com/articles/about-pull-request-merges/). I'm curious if that's a satisfactory equivalent.

mrwweb:
@grappler pretty sure I royally screwed up my branch with [this commit](https://github.com/Automattic/_s/pull/1017/commits/be200ef8cde9bc927c6f816b8a7fde3cffa0a70e) where I merged master (which I guess was unnecessary?). My branch now shows the entire history of commits since the original PR and so squashing the commits doesn't seem feasible. Should I just submit a new clean PR with a single commit?

grappler:
@mrwweb The PR is fixable so there is no need to create a PR. If you want to learn how to do it you can contact me on wordpress.slack.com or make sure the checkbox is selected to allow committers to modify your PR. I can the squash the commits and make sure you get the props for the changes.

grappler (approved):
Looks good to me. Can be merged.

PR #1023, opened 2016-09-20
Simplification of the _s_categorized_blog() return

alvisir:
Not a big change but a recommended simplification.

### DIFF
--- a/inc/template-tags.php
+++ b/inc/template-tags.php
@@ -98,7 +98,8 @@ function _s_categorized_blog() {
 
 		set_transient( '_s_categories', $all_the_cool_cats );
 	}
-
+	
+	// If this blog has more than 1 category the function return true, otherwise false.
 	return ( $all_the_cool_cats > 1 );
 }
 
--- a/inc/template-tags.php
+++ b/inc/template-tags.php
@@ -99,7 +99,7 @@ function _s_categorized_blog() {
 		set_transient( '_s_categories', $all_the_cool_cats );
 	}
 	
-	// If this blog has more than 1 category the function return true, otherwise false.
+	// If this blog has more than 1 category the function returns true, otherwise false.
 	return ( $all_the_cool_cats > 1 );
 }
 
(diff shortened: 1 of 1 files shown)


inc/template-tags.php
```diff
@@ -99,13 +99,7 @@ function _s_categorized_blog() {
...
 	}
 
-	if ( $all_the_cool_cats > 1 ) {
-		// This blog has more than 1 category so _s_categorized_blog should return true.
-		return true;
-	} else {
-		// This blog has only 1 category so _s_categorized_blog should return false.
-		return false;
-	}
+	return ( $all_the_cool_cats > 1 );
```
jrfnl: Maybe summarize the documentation which is now being removed and add it above the line ?

alvisir: You're right. Just updated the commit

inc/template-tags.php
```diff
@@ -98,14 +98,9 @@ function _s_categorized_blog() {
...
-
-	if ( $all_the_cool_cats > 1 ) {
-		// This blog has more than 1 category so _s_categorized_blog should return true.
-		return true;
-	} else {
-		// This blog has only 1 category so _s_categorized_blog should return false.
-		return false;
-	}
+	
+	// If this blog has more than 1 category the function return true, otherwise false.
```
jrfnl: `return` => `returns`

alvisir: Ouch! Thanks ;)

inc/template-tags.php
```diff
@@ -98,14 +98,9 @@ function _s_categorized_blog() {
...
-	if ( $all_the_cool_cats > 1 ) {
-		// This blog has more than 1 category so _s_categorized_blog should return true.
-		return true;
-	} else {
-		// This blog has only 1 category so _s_categorized_blog should return false.
-		return false;
-	}
+	
+	// If this blog has more than 1 category the function returns true, otherwise false.
+	return ( $all_the_cool_cats > 1 );
```
mtomas7: If we would also implement https://github.com/Automattic/_s/pull/1078/files

`return ( $all_the_cool_cats > 1 || is_preview() );`

davidakennedy:
We've removed this function as a result of #1131. Closing this out. Thanks for the work here @alvisir!

PR #1024, opened 2016-09-21
Remove Button Shadow Styling

joviczarko:
I would really like to hear your opinion on this. Thanks folks!

I think gradient styling of the buttons are unnecessary. They have been added long ago as background gradients, then changed in 2012 (#39 by @hugobaeta was merged on 14 Sep 2012 ). Now I think in the flat world we need to completely get rid of those gradients on the buttons.

Reasons for this:
- I assume everyone has it's own styling of the buttons
- I think people rarely uses that box shadow effect on buttons and text
- If we remove it we get flat buttons design. Which is much more used. 
- I remove it in every single project (most impotant reason :) )

But really, I think this styling only complicates _s theme unnecessarily.

Maybe we should add only this code to :hover pseudo class:
`background: lighten($color__background-button, 5%);`
To create more visible hover effect.

### DIFF
--- a/sass/forms/_buttons.scss
+++ b/sass/forms/_buttons.scss
@@ -6,21 +6,17 @@ input[type="submit"] {
 	border-color: $color__border-button;
 	border-radius: 3px;
 	background: $color__background-button;
-	box-shadow: inset 0 1px 0 rgba(255, 255, 255, 0.5), inset 0 15px 17px rgba(255, 255, 255, 0.5), inset 0 -5px 12px rgba(0, 0, 0, 0.05);
 	color: rgba(0, 0, 0, .8);
 	@include font-size(0.75);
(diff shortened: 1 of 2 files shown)


jrfnl:
Could you please add some screenshots showing the before and after ?

joviczarko:
Sure! Sorry for not making previews at the first place.

The Reset button is in hover state.

This is **BEFORE**
image

This is **AFTER**
image

Although the design is individual stuff, this shadow styled buttons now belongs to some past times. It is just waste of bytes. I think everyone is removing it on every project. And maybe it is time to be gone...

If we choose this path of simplifying some styling I can work on some other styling stuff to make things more universal.  For example making variables or mixins for borders to be consistent trough the site, to add a hover color variable, etc...

PR #1033, opened 2016-10-08
Unify widget title classes

youthkee:
In 404.php, there are two type of methods to include widgets.
Therefore there are different classes for widget titles.
`.widget-title` for "Most Used Categories" and `.widgettitle` for the others.
So I think it's better to unify these classes to `.widget-title`(popular class name nowadays).

### DIFF
--- a/404.php
+++ b/404.php
@@ -23,7 +23,7 @@ get_header(); ?>
 					<?php
 						get_search_form();
 
-						the_widget( 'WP_Widget_Recent_Posts' );
+						the_widget( 'WP_Widget_Recent_Posts', '', 'before_title=<h2 class="widget-title">' );
 
 						// Only show the widget if site has multiple categories.
 						if ( _s_categorized_blog() ) :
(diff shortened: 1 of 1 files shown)


karmatosed:
The classname comes from a core widget classname hence we have to have both.

grappler:
@karmatosed I think this ticket has merit and should be be reopened.

404.php
```diff
@@ -49,9 +49,9 @@
 
 						/* translators: %1$s: smiley */
 						$archive_content = '<p>' . sprintf( esc_html__( 'Try looking in the monthly archives. %1$s', '_s' ), convert_smilies( ':)' ) ) . '</p>';
-						the_widget( 'WP_Widget_Archives', 'dropdown=1', "after_title=</h2>$archive_content" );
+						the_widget( 'WP_Widget_Archives', 'dropdown=1', "before_title=<h2 class='widget-title'>&after_title=</h2>$archive_content" );
```
grappler: I would use an array for the args instead of the way it is currently done.

Issue #990, opened 2016-07-21
Site Title in the Customizer


samikeijonen:
I know that this has been discussed several times. Just want to check is there a small bug with the Customizer setting `Display Site Title and Tagline`.
1. Disable `Display Site Title and Tagline` under Site Identity.
2. Site Title and Tagline gets hidden in the Customizer view and when visiting site as it should.
3. Go away from the Customizer and visit the site.
4. Go back to Site Identity and enable `Display Site Title and Tagline`.
5. Now in the Customizer only Tagline will appear but not Site Title.

This is because in `inc/custom-header.php` there is this CSS:

```
.site-title,
.site-description {
    position: absolute;
    clip: rect(1px, 1px, 1px, 1px);
}
```

But in the `js/customizer.js` there is this JS with `site-title a`:

```
$( '.site-title a, .site-description' ).css( {
    'clip': 'rect(1px, 1px, 1px, 1px)',
    'position': 'absolute'
} );

...

$( '.site-title a, .site-description' ).css( {
    'clip': 'auto',
    'position': 'relative'
} );
```

I tested this JS without `a` tag using only `site-title` and that worked.

Am I missing something here or is the original behaviour as it should be?

davidakennedy:
I've run into this too. Will look at the PR.

PR #1034, opened 2016-10-16
Update .site-title styles instead of anchor in customizer script

joshmcrty:
See issue https://github.com/Automattic/_s/issues/990 from @samikeijonen 

Currently the customizer script visually hides/shows the `a` element inside the `.site-title`, which is not consistent with the actual CSS output by inc/custom-header.php, which hides/shows the `.site-title` element itself. This change makes the customizer consistent with the PHP changes.

### DIFF
--- a/js/customizer.js
+++ b/js/customizer.js
@@ -24,12 +24,12 @@
 	wp.customize( 'header_textcolor', function( value ) {
 		value.bind( function( to ) {
 			if ( 'blank' === to ) {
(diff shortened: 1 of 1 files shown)


mtomas7:
What about title color? See: https://github.com/Automattic/_s/pull/677

davidakennedy:
Tested this out, and it works nicely. It's also how I tweaked in some themes. Thanks @samikeijonen for the issue and @joshmcrty for the fix.

PR #1043, opened 2016-11-10
Security update: Remove automated Version numbers on scripts and styles

bahiirwa:
I have noticed that  several users of WordPress do not update their core on live production even when prompted too in the dashboard. Some security folks argue that this could be a problem when hackers use the version vulnerabilities for the "un-updated" core to hack the website. This information can easily be got from the source in the web browser 

Ben Sibley argues that for the meta tag generator versioning - "serves little purpose, but potentially is/could be used for gathering statistics by identifying sites as using WordPress. Some have said it is a potential security concern because it displays the version of WP you’re using. I’m no security expert, but I’m sure that’s an invalid claim or it wouldn’t be in core since V2.5."

Ref: https://www.competethemes.com/blog/wordpress-meta-tags-ultimate-guide/

I would love a second opinion on this. Thus the comment out in the functions.php file. If this workable, a merge would be good.

### DIFF
--- a/functions.php
+++ b/functions.php
@@ -99,6 +99,11 @@ function _s_widgets_init() {
 add_action( 'widgets_init', '_s_widgets_init' );
 
 /**
+ * Implement WordPres version stripping from theme styles & scripts.
+ */
+//require get_template_directory() . '/inc/version-strip.php';
+
+/**
  * Enqueue scripts and styles.
  */
 function _s_scripts() {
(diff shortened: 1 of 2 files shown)


grappler:
This is not something a theme should be doing. It creates a problem when switching themes and the next theme does not have this feature or I like the design of the theme but I don't like the extra things that the theme is doing. There are many Security Plugins that try to fix this, it would be better to install one of those.

There are so many other ways that a script can find the WP version of the site.

bahiirwa:
> The code you download from underscores.me is a snapshot of what we and contributing volunteers think is the best possible starting point for themes.

https://themeshaper.com/2013/10/11/dont-update-your-theme/

karmatosed:
I agree this isn't a theme side issue.

PR #1044, opened 2016-11-10
Simplify _s_categorized_blog function

mladenilic:


karmatosed:
I'm not sure this needs to happen as nobody has +1'd this, for now closing it.

Issue #1041, opened 2016-11-06
A11y issue: an empty link is formed when a custom header image is added


laviperchik:
In custom-header.php, line 9 the `alt` for the custom header image is empty. An empty `alt` is ok, accessibility-wise, as long as the `img` is not wrapped with a link. This is not the case here, and therefore an empty link is formed. 

I suggest adding an `alt` to the image, or adding the following code inside the link:
`<span class="screen-reader-text"><?php esc_html_e( 'homepage', 'underscores' ); ?></span>`

samikeijonen:
Thanks for noticing. I noticed the same in Core Custom logo feature. I'll ask around what would be best approach here.

samikeijonen:
What if we modify the example in couple of ways.

1. Should there really be link to home page. We already might have that link in logo, site title, and menu.
1. Let header image be decorative with empty `alt` like now.

Code:

```
	<?php if ( get_header_image() ) : ?>
		<img src="<?php header_image(); ?>" width="<?php echo esc_attr( get_custom_header()->width ); ?>" height="<?php echo esc_attr( get_custom_header()->height ); ?>" alt="">
	<?php endif; // End header image check. ?>
```

laviperchik:
I couldn't agree more with the solution you've suggested, header images are not clickable usually.

samikeijonen:
We should actually use [the_header_image_tag](https://developer.wordpress.org/reference/functions/the_header_image_tag/) function. It take cares of responsive markup. 

By default it actually has `get_bloginfo( 'name' )` as alt attribute.

samikeijonen:
Also noting that in WP 4.7 there will be function `the_custom_header_markup` which works for video headers also.

PR #1051, opened 2016-11-26
Remove link around header image

samikeijonen:
Fixes issue #1041.

### DIFF
--- a/footer.php
+++ b/footer.php
@@ -17,7 +17,7 @@
 		<div class="site-info">
 			<a href="<?php echo esc_url( __( 'https://wordpress.org/', '_s' ) ); ?>"><?php printf( esc_html__( 'Proudly powered by %s', '_s' ), 'WordPress' ); ?></a>
 			<span class="sep"> | </span>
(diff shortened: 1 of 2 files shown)


samikeijonen:
I switched to `the_header_image_tag()` function. It handles the responsive image markup by default.

grappler:
Really good change. Should be merged.

jrfnl:
I can't find any information on the minimum WP version `_s` is supposed to support at the moment (anyone?). As the `the_header_image_tag()` was introduced in WP 4.4, should the example code be surrounded by a `function_exists()` check ?

grappler:
`_s` supports the last 3 versions. That the what the TRT recommends and the `_s` follows. It should be fine if we wait till `4.7` is released.

jrfnl:
@grappler I just can't find any mention of that in the readme or anywhere else...

grappler:
@jrfnl You can find it mentioned in a few older PRs https://github.com/Automattic/_s/pull/644 The guideline is mentioned https://make.wordpress.org/themes/handbook/review/required/

Issue #1047, opened 2016-11-12
word-wrap in screen-reader-text class


samikeijonen:
In Core ticket [38764](https://core.trac.wordpress.org/ticket/38764) there is discussion about adding `word-wrap: normal` to `screen-reader-class` because screen readers may read out words ignoring spaces between words.

Even though `_s` uses `word-wrap: break-word` only in comment content link I can imagine that many themes add that style rule to body or content also. But they probably never change `screen-reader-class`.

Therefore it makes sense to add `word-wrap: normal` to `screen-reader-class`.

afercia:
For reference, see also:

WordPress core:
https://core.trac.wordpress.org/ticket/31962

Twenty Sixteen:
WordPress/twentysixteen@4a7578c

PR #1054, opened 2016-11-26
Add word-wrap: normal to screen-reader text class.

samikeijonen:
Fixes issue #1047.


pixelstrolch:
We should put this in the [sass file](https://github.com/Automattic/_s/blob/master/sass/modules/_accessibility.scss) as well.

PR #1055, opened 2016-12-18
Must use . to concatenate instead of ,

pratikshrestha:
In extras.php line 36, `.` needs to be used instead of `,` to concatenate  `bloginfo( 'pingback_url' )`


jrfnl:
No concatenation is being done on that line. It's an `echo` statement. An `echo` statements can take unlimited parameters separated by commas and it will echo each of them out in turn.
In other words: no concatenation needed. The benefit of this is that the memory consumed by PHP is lower when used like this (commas instead of concatenation).

Sugges: close.

PR #1058, opened 2016-12-21
add css load order fix

mithereal:
this function will allow the user to reorder their custom css placing it at the bottom of the array, so plugin styles etc dont override.

### DIFF
--- a/functions.php
+++ b/functions.php
@@ -115,6 +115,31 @@ function _s_scripts() {
 add_action( 'wp_enqueue_scripts', '_s_scripts' );
 
 /**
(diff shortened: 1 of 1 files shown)


grappler:
@mithereal Thank you for the PR. 

This will not be needed in 4.7.1 as the Custom CSS will loaded even later see: https://github.com/WordPress/WordPress/commit/15ddaa269e35baf0b721abb2d886f701550d59c5

When you contribute again in the future please make sure the function you add is prefixed with `_s_`. You can also see the issues that travis has reported on the coding styles https://travis-ci.org/Automattic/_s/jobs/185846669

karmatosed:
I feel with the addition of Custom CSS we can rule this out.

Issue #995, opened 2016-07-29
Selective Refresh on Site Title and Tagline


samikeijonen:
Should `_s` have selective refresh on Site Title and Tagline in the similar way as Twenty 16?

Here is [other reference](https://developer.wordpress.org/themes/advanced-topics/customizer-api/#selective-refresh-fast-accurate-updates).

celloexpressions:
Yes, and this is also required to show visible edit icons automatically in WordPress 4.7. Not difficult to add, but every theme needs it.

karmatosed:
Anyone up for a PR?

samikeijonen:
I can do PR in couple of days.

samikeijonen:
@celloexpressions: can you check is this PR ok?

PR #1063, opened 2016-12-28
Add selective refresh for site title and description.

samikeijonen:
This fixes issues #995.

### DIFF
--- a/inc/customizer.php
+++ b/inc/customizer.php
@@ -14,8 +14,37 @@ function _s_customize_register( $wp_customize ) {
 	$wp_customize->get_setting( 'blogname' )->transport         = 'postMessage';
 	$wp_customize->get_setting( 'blogdescription' )->transport  = 'postMessage';
 	$wp_customize->get_setting( 'header_textcolor' )->transport = 'postMessage';
+
+	if ( isset( $wp_customize->selective_refresh ) ) {
+		$wp_customize->selective_refresh->add_partial( 'blogname', array(
+			'selector'        => '.site-title a',
+			'render_callback' => '_s_customize_partial_blogname',
+		) );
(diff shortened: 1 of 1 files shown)


inc/customizer.php
```diff
@@ -14,10 +14,39 @@ function _s_customize_register( $wp_customize ) {
 	$wp_customize->get_setting( 'blogname' )->transport         = 'postMessage';
 	$wp_customize->get_setting( 'blogdescription' )->transport  = 'postMessage';
 	$wp_customize->get_setting( 'header_textcolor' )->transport = 'postMessage';
+
+	if ( isset( $wp_customize->selective_refresh ) ) {
+		$wp_customize->selective_refresh->add_partial( 'blogname', array(
+			'selector'        => '.site-title a',
+			'render_callback' => '_s_customize_partial_blogname',
```
grappler: Looking at the documentation it would be 

```php
function() {
      return get_bloginfo( 'name', 'display' );
}
```

https://make.wordpress.org/core/2016/02/16/selective-refresh-in-the-customizer/

samikeijonen: Isn't `bloginfo( 'name' )` the same as `return get_bloginfo( 'name', 'display' )`?

davidakennedy:
@samikeijonen Want to refresh this PR? I'd love to get it in ASAP. Seems  like you can just following the example set in Twenty Seventeen.

grappler:
@davidakennedy I refreshed for @samikeijonen I hope that is Ok for both of you and then we can merge this.

samikeijonen:
I'm OK with that, thanks!

davidakennedy:
Me too! Thanks @grappler and @samikeijonen!

PR #1066, opened 2017-01-03
Prevent broken translations in some locales for the edit post link 

yoavf:
Since #775, the translation of the _edit post link_ in some languages can end up forming a partial sentence.

By default, in English, the visual output is the word "**Edit**".
In Hebrew for example, the translation would be a single word:
<pre><p dir='rtl' align='right'>לערוך</p></pre>

With the addition of the post title, the translator is instructed to translate `Edit %s` with `%s` being the post name. In Hebrew, the translation will be
<pre><p dir='rtl' align='right'>לערוך את %s</p></pre>
Note the added word `את`

To quote from [this post](https://myhebrewwords.wordpress.com/2014/03/07/1-%D7%90%D7%AA-et-the-most-common-word-in-the-hebrew-language/):

> The word **את** comes before a definite object. It has no equivalent in the English language.

The end visual result in Hebrew would be:
<pre><p dir='rtl' align='right'>לערוך את</p></pre>

Which is weird. Think of it as "Edit The" (without the actual object).

This PR fixes that by expanding the comment to note that the post name only visible to screen readers, and puts the screen reader span html inside the translated string.
The translator for Hebrew can now do something like
`לערוך <span class="screen-reader-text">את %s</span>`, putting the את word inside the invisible span.

### DIFF
--- a/inc/template-tags.php
+++ b/inc/template-tags.php
@@ -68,7 +68,7 @@ function _s_entry_footer() {
 
 	edit_post_link(
 		sprintf(
-			/* translators: %s: Name of current post */
+			/* translators: %s: Name of current post. Only visible to screen readers */
 			__( 'Edit <span class="screen-reader-text">%s</span>', '_s' ),
 			get_the_title()
 		),
--- a/inc/template-tags.php
+++ b/inc/template-tags.php
@@ -69,7 +69,7 @@ function _s_entry_footer() {
 	edit_post_link(
 		sprintf(
 			/* translators: %s: Name of current post. Only visible to screen readers */
-			__( 'Edit <span class="screen-reader-text">%s</span>', '_s' ),
+			wp_kses( __( 'Edit <span class="screen-reader-text">%s</span>', '_s' ), array( 'span' => array( 'class' => array() ) ) ),
 			get_the_title()
 		),
 		'<span class="edit-link">',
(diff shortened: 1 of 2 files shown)


inc/template-tags.php
```diff
@@ -68,9 +68,9 @@ function _s_entry_footer() {
 
 	edit_post_link(
 		sprintf(
-			/* translators: %s: Name of current post */
-			esc_html__( 'Edit %s', '_s' ),
-			the_title( '<span class="screen-reader-text">"', '"</span>', false )
+			/* translators: %s: Name of current post. Only visible to screen readers */
+			__( 'Edit <span class="screen-reader-text">%s</span>', '_s' ),
```
grappler: Previously the function was `esc_html__()`. It is best not to change that.

yoavf: That's won't work now that we have the html back in the string.

grappler: Oh, then we will need to use `wp_kses()` like done here: https://github.com/Automattic/_s/blob/master/inc/template-tags.php#L65

yoavf:
>  believe this could have been fixed by just adjusting the translators comment to be more clear, rather than changing the actual code.

My first instinct was to do just that - but that means (for Hebrew) choosing to omit a word from the screen reader version. It is not a better outcome.

davidakennedy:
Looks good! Thanks @yoavf @grappler and @jrfnl!

PR #1070, opened 2017-01-09
Matching style.css

bahiirwa:
Adding word-wrap to .screen-reader-text
* Many screen reader and browser combinations announce broken words as they would appear visually.


bahiirwa:
Ref: #1047

PR #1078, opened 2017-01-31
Allow category display in post preview even when only one category

mtomas7:
Via: https://core.trac.wordpress.org/changeset/40023


davidakennedy:
Looks good @mtomas7! Thanks!

PR #1090, opened 2017-03-02
Update _menus.scss

kkalvaa:
Change 1.5em to 100% to prevent magic numbers.

### DIFF
--- a/sass/navigation/_menus.scss
+++ b/sass/navigation/_menus.scss
@@ -14,7 +14,7 @@
 			box-shadow: 0 3px 3px rgba(0, 0, 0, 0.2);
 			float: left;
 			position: absolute;
-			top: 1.5em;
+			top: 100%;
 			left: -999em;
 			z-index: 99999;
 
--- a/style.css
+++ b/style.css
@@ -559,7 +559,7 @@ a:active {
 	box-shadow: 0 3px 3px rgba(0, 0, 0, 0.2);
 	float: left;
 	position: absolute;
-	top: 1.5em;
+	top: 100%;
 	left: -999em;
 	z-index: 99999;
 }


davidakennedy:
Hi @kkalvaa and thanks for the pull request! Welcome to _s!

Could you explain you reasoning for this change more? I'm not sure what "magic numbers" means in this case. If the current issue is causing some visual issues, it would be good to have a screenshot to see it.

kkalvaa:
Hi @davidakennedy 

[Magic numbers](https://css-tricks.com/magic-numbers-in-css/) is a concept referring to overly specific numbers, in this case 1.5em. The reason it works is because the size of the box is 1.5em, however once you introduce padding or borders, the box itself will be larger. Since the drop down is then moved down 1.5em it will now overlap the parent box, rather than sit flush with the box as it used to. Or if you change the font size the box will be a different size leading 1.5em from the top to again be possibly problematic. By changing to 100% it will always be flush with the parent box regardless of the size.

The point is, top: 1.5em; is works because everything else is as it should. top: 100%; works regardless of what styling is applied to the menu.

davidakennedy:
@kkalvaa Thanks for the explanation! Makes a lot more sense, and I think it's a good change to make.

Can you please update the main stylesheet as well? Per our [contribution guidelines](https://github.com/Automattic/_s/blob/master/CONTRIBUTING.md), "When submitting a change that affects CSS, please make sure that both SCSS sources and output CSS have been updated equally."

We don't have an automated way to compile on commit currently. Thanks!

kkalvaa:
Here's where I show how new I am to GitHub. How would I go about to add a file to this commit?

davidakennedy:
@kkalvaa No worries! You committed changes to your `patch-1` branch. If you commit new changes to the same branch, and push the changes to your fork, it will update this pull request.

See this [help article for more](https://help.github.com/articles/about-pull-requests/).

kkalvaa:
Thanks for the info, I've gone ahead and done just that.

kkalvaa:
I'm not sure what you mean. That's the original file I changed, and then added the same change in to style.css.

davidakennedy:
Thanks @kkalvaa! 😄

PR #1103, opened 2017-04-27
Update header.php

maperezotero:
Using And && instead of OR ||, the h1 tag will be used only if we select to display our latest posts in the front page. If we choose a static page as our front page, the h1 tag will never be used.
Using OR ||, the h1 tag will be used in the front page and in the posts page.

### DIFF
--- a/header.php
+++ b/header.php
@@ -26,7 +26,7 @@
 	<header id="masthead" class="site-header" role="banner">
 		<div class="site-branding">
 			<?php
-			if ( is_front_page() && is_home() ) : ?>
+			if ( is_front_page() || is_home() ) : ?>
 				<h1 class="site-title"><a href="<?php echo esc_url( home_url( '/' ) ); ?>" rel="home"><?php bloginfo( 'name' ); ?></a></h1>
 			<?php else : ?>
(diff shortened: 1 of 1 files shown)


iliman:
> If we choose a static page as our front page, the h1 tag will never be used.

If we choose a static page as our front page, h1 will be used for the page title.
And if you do as you suggest, in this case there will be two h1 tags on the page, which is incorrect.

mtomas7:
> two h1 tags on the page, which is incorrect.
From the HTML5 perspective, it would be correct, but we could say, that it would be undesirable from the SEO perspective.

davidakennedy:
Thanks for the pull request @maperezotero and welcome!

It's this way for a reason. See: https://github.com/Automattic/_s/commit/dc6d3fa08ca3669081ebac794d6046b385088552

Currently, we always have a nice hierarchy, with one `h1` on the page. Even though it's valid in HTML5 to have more than one `h1`, it's not useful to [accessibility](https://www.paciellogroup.com/blog/2013/10/html5-document-outline/) or SEO. If you've set a static page as your front page, it makes sense that in most cases you'd want that page's title to be one, true `h1`. not your site title.

As with most things related to accessibility and SEO, some probably have other opinions. 😄 

We could change this up to make the page title of a static page as front page a `h2` with the site title a `h1`. But I'm not sure there's a strong reason to do that.

I'll close this for now, but if strong reasons surface, we can always revisit.

PR #1105, opened 2017-05-09
Fix latest phpcs update

AngeIII:


jrfnl:
This is not going to work (yet) as the Wordpres Coding Standard is not (yet) compatible with PHPCS 3.x
See: https://github.com/Automattic/_s/blob/master/.travis.yml

For now, you'd need to change the PHPCS branch which is being cloned to `2.9`.

AngeIII:
Confirm that, for now you can change travis script to `    - if [[ "$SNIFF" == "1" ]]; then git clone -b 2.8.1 --depth 1 https://github.com/squizlabs/PHP_CodeSniffer.git $PHPCS_DIR; fi
`
then it will work fine with latest wp standards.

jrfnl:
@AngeIII Actually it won't. This repo clones the `master` branch of WPCS which currently has a minimum PHPCS requirement of version `2.9.0`.
If you use the latest tagged release of WPCS, then yes, `2.8.1` is fine, but that's not the case here.

davidakennedy:
I've switched the branch on PHPCS to `2.9` so things work for now until WPCS is compatible. Thanks for bringing this up!

Issue #1064, opened 2016-12-28
Rename extras.php


grappler:
In the header stands
```php
 * Custom functions that act independently of the theme templates
 *
 * Eventually, some of the functionality here could be replaced by core features.
```

There are two functions in the file `_s_body_classes()` & `_s_pingback_header()`. I don't think they will be replaced by core functions.

I find that the `extras.php` becomes a dumping ground for all functions that don't fit in the other files when developing themes.

In Twenty Seventeen the [`twentyseventeen_body_classes()`](https://github.com/WordPress/twentyseventeen/blob/master/inc/template-functions.php#L16) is in [`template-functions.php`](https://github.com/WordPress/twentyseventeen/blob/master/inc/template-functions.php) and `twentyseventeen_pingback_header()` is placed in `functions.php`

davidakennedy:
Thanks @grappler! Probably a good idea to follow Twenty Seventeen's lead and rename it ` template-functions.php` to make it more clear as to the purpose. We could also change the header description. PRs welcome!

jrfnl:
> In Twenty Seventeen the twentyseventeen_body_classes() is in template-functions.php and twentyseventeen_pingback_header() is placed in functions.php

The fact that Twenty Seventeen has moved the `pingback_header()` function to the `functions.php` file should not be a reason to do the same here.

To me it feels more logical to keep the `pingback_header()` function in the `template-functions.php` file as it adds HTML directly to the page. None of the other functions in `functions.php` do this.

I'd like to suggest reopening this issue and discussing this move more thoroughly.

davidakennedy:
> To me it feels more logical to keep the pingback_header() function in the template-functions.php file as it adds HTML directly to the page. None of the other functions in functions.php do this.

@jrfnl That's a good point. I could go either way. Happy to see a PR here with the change. 😄

jrfnl:
For the PR, see #1118

PR #1111, opened 2017-06-09
Replace extras.php with template-functions.php

grappler:
@davidakennedy Here is the PR for #1064


davidakennedy:
Looks good @grappler! Thanks!

PR #1117, opened 2017-06-12
Improve "Continue reading" word order for i18n

### LINKED ISSUE #1088: Conflict "screen-reader-text" and i18n word order change on continue reading link
I saw hidden post title added to the continue reading link since https://github.com/Automattic/_s/pull/516.

But in Japanese language, we have different translation for `continue reading` and `continue reading %s`. We should insert particle `の` between object and `continue reading`. If string doesn't have object, we should remove `の`。

Original|Japanese
--|--
continue reading|続きを読む
continue reading %s|%s**の**続きを読む

As result, we can see only `の続きを読む` without object with display and it doesn't make sense. Is there any better solution?

See also: [Wikipedia: Japanese particle: no](https://en.wikipedia.org/wiki/Japanese_particles#no)

# My suggestion
## A: update translator comment and translate as separate sentence
In default themes like Twenty Fifteen, we translate it as separated sentence. but [most theme translator doesn't know this situation](https://translate.wordpress.org/consistency?search=Continue+reading+%25s&set=ja%2Fdefault). I want to tell them the post name will be removed on screen. This is not the best solution because `Continue reading. Post name.` doesn't form a sentence.

## B: add different translation
I prefer it, but how?

naokomc:
Resolves translation word order issue for some locales when the screen reader only text is invisible.
Fixes #1088

### DIFF
--- a/template-parts/content.php
+++ b/template-parts/content.php
@@ -30,15 +30,15 @@
 		<?php
 			the_content( sprintf(
 				wp_kses(
-					/* translators: %s: Name of current post. */
-					__( 'Continue reading %s <span class="meta-nav">&rarr;</span>', '_s' ),
+					/* translators: %s: Name of current post. Only visible to screen readers */
+					__( 'Continue reading<span class="screen-reader-text"> "%s"</span>', '_s' ),
 					array(
 						'span' => array(
 							'class' => array(),
 						),
 					)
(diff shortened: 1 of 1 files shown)


davidakennedy:
Thanks for the pull request @naokomc!

PR #1118, opened 2017-06-12
Move pingback function back to the `template-functions` file.

jrfnl:
As discussed in #1064

### DIFF
--- a/inc/template-functions.php
+++ b/inc/template-functions.php
@@ -1,6 +1,6 @@
 <?php
 /**
- * Additional features to allow styling of the templates
+ * Functions which enhance the theme by hooking into WordPress
  *
  * @package _s
  */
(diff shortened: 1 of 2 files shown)


inc/template-functions.php
```diff
@@ -1,6 +1,6 @@
 <?php
 /**
- * Additional features to allow styling of the templates
+ * Functions which enhance the content displayed by hooking into WP
```
davidakennedy: Let's not abbreviate WordPress here, just so it's clear. Also, maybe change "the content" to "the theme" since what's changed could extend beyond actual content.

jrfnl: Good point. Fixed up the commit.

davidakennedy:
Thanks @jrfnl for the pull request! 😄

PR #1120, opened 2017-06-15
Consistent formatting for skip-link-focus-fix.js

JulienMelissas:
Just noticed the spacing on the other files and not this one... See https://github.com/Automattic/_s/blob/master/js/customizer.js#L9 and https://github.com/Automattic/_s/blob/master/js/customizer.js#L42


davidakennedy:
Good catch @JulienMelissas! Thanks for the pull request!

PR #1121, opened 2017-06-15
PHPCS ruleset: Improve documentation links

jrfnl:
#### Fix link to WPCS in the PHPCS ruleset.

The link was pointing to the `WordPress-Core` ruleset, while `_s` uses the `WordPress` ruleset which encompasses more than just the core rules.

Pointing to the repo + the wiki homepage will give people a better starting point.

#### Add link to the PHPCompatibility repo.

At the top of the ruleset there were links to the PHPCS and the WPCS repo, but not to the PHPCompatibility repo which is also being used.


davidakennedy:
Thanks @jrfnl!

PR #1122, opened 2017-06-15
WPCS: Allow for typical theme specific file names based on the theme hierarchy

jrfnl:
WPCS checks that words in file names are separated by dashes, not underscores.

For themes, a number of exceptions are made to this rule, to allow for `taxonomy-{$taxo_nomy}.php` and `archive-{$post_type}.php` type of file names.

While `_s` itself currently does not contain any files which would require the exception, it is likely that derivative themes will have these kind of files.
Setting the property in the PHPCS ruleset used for `_s` will thereby give a good example of how this should be handled (and prevent people from excluding the sniff completely).

Refs:
* https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/wiki/Customizable-sniff-properties#themes-allow-filename-exceptions
* https://developer.wordpress.org/themes/basics/template-hierarchy/
* http://wphierarchy.com/

### DIFF
--- a/phpcs.xml.dist
+++ b/phpcs.xml.dist
@@ -36,6 +36,14 @@
 		</properties>
 	</rule>
 
+	<!-- Allow for theme specific exceptions to the file name rules based
+		 on the theme hierarchy. -->
+	<rule ref="WordPress.Files.FileName">
+		<properties>
+			<property name="is_theme" value="true" />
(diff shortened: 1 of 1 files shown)


davidakennedy:
Thanks for the pull request @jrfnl!

PR #1125, opened 2017-06-17
WPCS: Verify that no deprecated or removed WP functions are being used.

jrfnl:
Since WPCS 0.11.0, it is possible to check that no WP deprecated functions are used.

By default, the minimum wp version against which this is checked is 3 versions before the current release (would be 4.5 now as 4.8 is out), however, the minimum wp version used is configurable and should by rights be in line with the minimum WP version for the theme as proclaimed in the `readme.txt` file.
This minor change sorts that out as well as informs theme-devs about this feature.

Ref: https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/wiki/Customizable-sniff-properties#minimum-wp-version-to-check-for-usage-of-deprecated-functions-and-function-parameters

### DIFF
--- a/phpcs.xml.dist
+++ b/phpcs.xml.dist
@@ -45,6 +45,15 @@
 		</properties>
 	</rule>
 
+	<!-- Verify that no WP functions are used which are deprecated or have been removed.
+		 The minimum version set here should be in line with the minimum WP version
+		 as set in the "Requires at least" tag in the readme.txt file. -->
+	<rule ref="WordPress.WP.DeprecatedFunctions">
+		<properties>
+			<property name="minimum_supported_version" value="4.0" />
+		</properties>
+	</rule>
+
 	<!-- Include sniffs for PHP cross-version compatibility. -->
 	<config name="testVersion" value="5.2-99.0"/>
 	<rule ref="PHPCompatibility"/>


phpcs.xml.dist
```diff
@@ -45,6 +45,15 @@
 		</properties>
 	</rule>
 
+	<!-- Verify that no WP functions are used which are deprecated or have been removed.
+		 The minimum version set here should be in line with the minimum WP version
+		 as set in the "Requires at least" tag in the readme.txt file. -->
+	<rule ref="WordPress.WP.DeprecatedFunctions">
+		<properties>
+			<property name="minimum_supported_version" value="4.0" />
```
grappler: Is there a reason why you used `4.0` instead of a higher version?

jrfnl: Yes, to quote myself:
_The minimum version set here should be in line with the minimum WP version as set in the "Requires at least" tag in the readme.txt file._

And the `readme.txt` file says `4.0` at the moment.
https://github.com/Automattic/_s/blob/master/readme.txt#L6

grappler: Ah, sorry, I misunderstood.

jrfnl: Happens to the best of us ;-)

davidakennedy:
This looks good. Thanks for the continued improvements here @jrfnl!

PR #1126, opened 2017-06-17
Clean up comment

bahiirwa:
Add the hint properly in the comments section

### DIFF
--- a/inc/custom-logo.php
+++ b/inc/custom-logo.php
@@ -11,8 +11,9 @@
  * Get the custom Logo attachment link
  *
  */
-function _s_the_custom_logo() {
-	if ( function_exists( '_s_the_custom_logo' ) ) {
+
+if ( ! function_exists( '_s_the_custom_logo' ) ) {
+	function _s_the_custom_logo() {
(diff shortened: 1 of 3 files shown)


inc/custom-logo.php
```diff
@@ -0,0 +1,19 @@
...
+ * Sample implementation of the Custom Logo feature
+ *
+ * <?php _s_the_custom_logo(); ?>
+ *
+ * @package _s
+ */
+
+/**
+ * Get the custom Logo attachment link
+ *
```
jrfnl: Superfluous empty comment line

inc/custom-logo.php
```diff
@@ -0,0 +1,19 @@
...
+ */
+function _s_the_custom_logo() {
+	if ( function_exists( '_s_the_custom_logo' ) ) {
```
jrfnl: Calling the `function_exists()` from within the function it checks for, however, will always return `true`.

I think you may have meant to wrap the whole function declaration within a `if ( ! function_exists( '_s_the_custom_logo' ) ) {}` wrapper ?

Issue #1019, opened 2016-09-16
Add editor-style.css to _s?


mrwweb:
The editor-style.css file is far and away one of the best ways to improve a site editor's experience (at least in my experience). It makes TinyMCE much WYSIWYGier(TM) and helps users understand the connection between a theme and typography styles.

So in that vein, I'd love to see _s add editor-style.css by default to the theme. Thinking through this, I believe that would involve doing the following:
1. Add `add_editor_style();` to `functions.php`
2. Add an `editor-style.css` file to the theme's root and include a brief description and link to documentation in a comment.
3. Add an `editor-style.scss` file to the `/sass/` folder that includes the same comments as above AND `@import`s helpful partials including:
   - everything in `/sass/typography/`
   - everything in `/sass/elements/`
   - everything in `/sass/modules/`
   - `normalize`
   - all requires mixins and variables partials to make the above compile

I've taken the approach in #3 with my own themes on multiple sites and have found a lot of success in autogenerating a nearly-complete set of editor styles. I don't know the _s SASS setup that well, so some input on that would be useful.

I thought a discussion would be worthwhile before submitting a pull request, so please let me know what you think. As far as I can tell, this hasn't been considered before.

josephfusco:
Looks like this was considered before but closed on the premise that there are no styles in this starter theme.

There was however no sass implementation when [this PR ](https://github.com/Automattic/_s/pull/225) was closed and believe that laying out the sass structure you mentioned could be beneficial as it is not adding in any new styles. I personally use approach 3 you mentioned and always end up manually creating `editor-style.scss` importing all partials related to themes post styles.

mrwweb:
@josephfusco Glad to hear that strategy works for more than just me. Can you find the previous discussions? My initial searching didn't turn up anything.

In theory, I guess a pull could include the same styles in the non-SASS `style.css`, my only concern there is that it would add a third place that CSS needs to be updated frequently which feels like a losing proposition. I'm also seen people just include their full-site stylesheet as the editor style, but I don't think that works well in practice (it's bloated, and styles often assume markup that isn't in the editor, such as styling `.site-content h2`).

And just to look at other precedents, every Twenty XXXX default WordPress theme has shipped with an `editor-style.css`, so this is a clearly a core-blessed best practice.

josephfusco:
Ref: https://github.com/Automattic/_s/pull/225

valleyspirit:
[...]

PR #1130, opened 2017-06-22
Add editor-style.css to educate themers

mrwweb:
First pass on a PR for #1019. It does the following:

- Adds `editor-style.css` and `editor-style.scss`
- Loads the styles in `functions.php`
- Updates `CONTRIBUTING.md` to reflect these changes

#1019 has a long discussion on the best `_s`-appropriate approach to editor styles, and this PR is my attempt at the most robust version of that. In short, the changes introduce:

- no new CSS rules to manage (not including `@import`s which generate styles but don't require additional maintenance barring a major project-wide refactor),
- a solution similar to the existing `rtl.css` file, and
- educational comments with links to relevant reading.

The goal is to educate developers on the benefits of editor styles and not let people forget them in a way that is efficient but doesn't dictate the styles themselves.

Props to @josephfusco for the starting point to the SCSS file.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -13,7 +13,7 @@ Here are some tips to consider and to help you write a great report:
 * `_s` supports Microsoft Internet Explorer 11 and Edge, as well as the latest two versions of all other major browsers.
 * `_s` is backwards compatible with the two versions prior to the current stable version of WordPress.
 * `_s` uses HTML5 markup.
-* We decided not to include translations [[#50](https://github.com/Automattic/_s/pull/50)] beyond the existing `_s.pot` file, a RTL stylesheet [[#263](https://github.com/Automattic/_s/pull/263)], or editor styles [[#225](https://github.com/Automattic/_s/pull/225)], as they are likely to change during development of an `_s` based theme.
+* We decided not to include translations [[#50](https://github.com/Automattic/_s/pull/50)] beyond the existing `_s.pot` file, RTL CSS styles [[#263](https://github.com/Automattic/_s/pull/263)], or editor CSS styles [[#1019](https://github.com/Automattic/_s/issues/1019)], as they are likely to change during development of an `_s` based theme.
 
 ## Contributing code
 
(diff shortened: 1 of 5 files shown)


davidakennedy:
Thanks for the pull request @mrwweb!

Overall, looks pretty good. Here's are some thoughts and feedback:

* You have some whitespace at the end of line 71 of `functions.php` that's causing the build to fail. Easy fix. 😄 
* Why do you import some things in the `.scss` file, but those aren't reflected in the final stylesheet? I think my preference here would be to go lighter. Include just the stuff under "Custom Editor Styles" as a starting point and be done with it. Keep it simple. But others may have different opinions.

davidakennedy:
I'm going to close this out since editor styles will become a thing of the past once Gutenberg hits. See: https://github.com/Automattic/_s/issues/1019#issuecomment-310682439

Thanks for the pull request, and solid code! 😄

mrwweb:
Responding to @davidakennedy on the off chance things change upstream and this is worth resurrecting.

> Why do you import some things in the .scss file, but those aren't reflected in the final stylesheet?

While the SASS version updates itself based on changes to the front end theme, the CSS version would have to pull in a lot of styles likely to change. At that point, the themer either has to manually keep style.css and editor-style.css in sync or just start over in editor-style.css, selectively pulling in the correct styles.

My experience with both SASS and non-SASS editor styles seemed to align with others in the discussion on #1019, so I think the files setup as they were encouraged the most appropriate workflow for both versions.

And as mentioned in the opening comment, both ways introduce files so that there are no additional CSS rules that would need to be changed when front-end style.css changes are made to the project.

davidakennedy:
Reopening to continue work here based on feedback in this comment: https://github.com/Automattic/_s/issues/1019#issuecomment-311499717

mrwweb:
@davidakennedy Responding to the comment you just made on the issue since it regards this PR specifically:

> I think a good compromise is to just add an empty file with some very basic styles, a la the rtl.css file.

Are you specifically proposing the removal of the `@import`s from the SASS version? If yes, I'd at least want to include a SASS-specific comment suggesting the technique.

davidakennedy:
Closing this, as I also closed #1019.

PR #1131, opened 2017-06-23
_s: Add basic support for Content Options Post Details

thomasguillot:
This allows us to always display the category and author since it can
be hidden via Customizer -> Content Options

We don’t need to check for `is_multi_author` or `_s_categorized_blog()`
anymore

### DIFF
--- a/inc/jetpack.php
+++ b/inc/jetpack.php
@@ -27,14 +27,14 @@ function _s_jetpack_setup() {
 
 	// Add theme support for Content Options.
 	add_theme_support( 'jetpack-content-options', array(
-	    'post-details'       => array(
-	        'stylesheet'      => '_s-style',
-	        'date'            => '.posted-on',
-	        'categories'      => '.cat-links',
-	        'tags'            => '.tags-links',
-	        'author'          => '.byline',
-	        'comment'         => '.comments-link',
-	    ),
+		'post-details'       => array(
+			'stylesheet'      => '_s-style',
+			'date'            => '.posted-on',
+			'categories'      => '.cat-links',
+			'tags'            => '.tags-links',
+			'author'          => '.byline',
+			'comment'         => '.comments-link',
+			),
 	) );
 }
 add_action( 'after_setup_theme', '_s_jetpack_setup' );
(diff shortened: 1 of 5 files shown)


davidakennedy:
Thanks for the pull request @thomasguillot!

This looks good. I made a few tweaks for spacing because of a failed build.

Beyond that, everything works as it should.

A few other notes of feedback:

* We talked about this a bit via Slack: With Content Options on the whole, we should remove the items with `display: none;` vs. hiding things off screen. Because this is really a design decision. And I can't think of a reason why these items should be hidden only visually. There's no needed context for accessibility or SEO that I can think of.
* We should add featured image support here and to _s. It would be a good way to standardize the use of featured images across themes. And hiding them is something we know customers want.

davidakennedy:
Thinking about this more, it makes sense to merge this now, and we can do a featured image addition in another pull request. Since that will be a good bit of code, it makes sense to separate out things.

Thanks @thomasguillot!

jrfnl:
I'm seriously wondering why this got merged as it is.

I know that on .com JetPack is enabled by default, but this is a starter theme which is also used by numerous theme authors to create themes and when one of those themes is used, the chance that JetPack is enabled goes down significantly.

Please don't remove good functionality in favour of demanding the use of an arbitrary plugin.

davidakennedy:
Thanks for the comment @jrfnl! That's good feedback, and it's always welcome.

This doesn't change anything significant in regards to Jetpack. If authors don't want to use it, they can remove the file and the associated include in `functions.php` – just as before. Or delete the parts of Jetpack that aren't needed.

Underscores is the project that the Automattic Theme Team uses to create themes for WordPress.com. So sometimes, things will go in because they make sense for the majority of WordPress.com customers. Whenever that happens, myself and the rest of the maintainers try to make it easy to remove or adjust anything that community theme authors might not need.

jrfnl:
@davidakennedy I understand that and I'm absolutely not against the addition which was made.

My remark is aimed at the fact that this PR also **removed** existing functionality which was useful for "the rest of us".

davidakennedy:
@jrfnl Totally understand. We probably wouldn't have yanked out those functions, but they interfered with Content Options. Also, many WordPress.com customers found it confusing that themes didn't show categories unless a site had more than one. It was inconsistent with their expectations, and we're really after making everything as consistent as possible.

Hope this helps clarify things a bit. 😄

PR #1133, opened 2017-06-23
Remove landmark roles

### LINKED ISSUE #1132: Remove Landmark Roles
With WordPress moving ahead with [better browser support](https://make.wordpress.org/core/2017/04/23/target-browser-coverage/), and _s [supporting Microsoft Internet Explorer 11 and Edge, as well as the latest two versions of all other major browsers](https://github.com/Automattic/_s/blob/master/CONTRIBUTING.md), removing the landmarks roles makes sense.

They're implied with the proper HTML5 markup, and warnings are given in most code checking tools if they're added. For more information, see [this post from 2015](http://html5doctor.com/on-html-belts-and-aria-braces/) by  Steve Faulkner that recommends dropping them.

We've kept them for quite awhile to have better support with assistive technology, but it's time to move forward.

I [brought this up](https://wordpress.slack.com/archives/C02RP4X03/p1497892561177146) with the WordPress Accessibility team, and they're for it.

Previously:

#786 
#797 
#889 

and probably more. 😄

davidakennedy:
* With better browser support, these roles aren't needed now.
* Most newer browsers imply them via HTML5 semantics. See: http://html5doctor.com/on-html-belts-and-aria-braces/

Fixes #1132

### DIFF
--- a/404.php
+++ b/404.php
@@ -10,7 +10,7 @@
 get_header(); ?>
 
 	<div id="primary" class="content-area">
-		<main id="main" class="site-main" role="main">
+		<main id="main" class="site-main">
 
 			<section class="error-404 not-found">
 				<header class="page-header">
--- a/archive.php
+++ b/archive.php
@@ -10,7 +10,7 @@
 get_header(); ?>
 
 	<div id="primary" class="content-area">
-		<main id="main" class="site-main" role="main">
+		<main id="main" class="site-main">
 
 		<?php
 		if ( have_posts() ) : ?>
--- a/comments.php
+++ b/comments.php
@@ -46,7 +46,7 @@ if ( post_password_required() ) {
 		</h2><!-- .comments-title -->
 
 		<?php if ( get_comment_pages_count() > 1 && get_option( 'page_comments' ) ) : // Are there comments to navigate through? ?>
-		<nav id="comment-nav-above" class="navigation comment-navigation" role="navigation">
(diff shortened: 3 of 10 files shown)


pixelstrolch:
A smaller markup is great, however I'd like to reference to the statement on the official a11y tutorials page: 
>Most current web browsers support the above HTML5 elements and convey the information to assistive technology through the accessibility APIs. However, to maximize compatibility with web browsers and assistive technologies that support WAI-ARIA but do not yet support HTML5, it is currently advisable to use both the HTML5 elements and the corresponding WAI-ARIA roles.
https://www.w3.org/WAI/tutorials/page-structure/regions/#accessupport

If we want to support IE11, I think we should keep the main ARIA landmark.
http://www.html5accessibility.com/

Also the [a11y handbook](https://make.wordpress.org/accessibility/handbook/quick-start-guide/#landmarks) doesn't seem to be ready for this change. This might cause some discussions in the review process.

davidakennedy:
Thanks for the thoughts @pixelstrolch!

Myself and the WordPress Accessibility feel like the tradeoff for one browser at this point is fine in order to have more future-friendly markup. Besides, if you run a site through the official [W3C validator](https://validator.w3.org/) with a defined role on an corresponding HTML5 element, like `<header>` for example, it throws a warning. Also, we kept these around for a long time. 😄

PR #1137, opened 2017-06-27
Add partial refresh for site title and tagline in Customizer

### LINKED ISSUE #995: Selective Refresh on Site Title and Tagline

ernilambar:
Fixes #995 

* Existence of partial refresh is not checked as it is available since WP 4.5


grappler:
Duplicate of #995

Issue #1116, opened 2017-06-12
Set up all functions to be overloadable


jrfnl:
At this moment, only the `_s_posted_on()` and `_s_entry_footer()` functions in `template-tags.php` and `_s_setup()` in `functions.php` are set up to allow for overloading by a potential child theme.

While `_s` is intended to be a base theme and not a parent theme, this does not negate the possibility that a derivative theme based on `_s` would become a parent-theme for other child-themes.

To that end, I'd like to suggest making all methods within `_s` overloadable by a child-theme, i.e. wrapping all functions within a `if ( ! function_exists( 'function_name' ) ) {}` wrapper.

Opinions ?

grappler:
We should not use pluggable functions if they are being used in hooks as we can use `remove_action/filter` in the child theme. Template files are the only ones that we cannot remove and need to be overridden.

Related: https://github.com/Automattic/_s/issues/929 & https://github.com/WordPress/twentyseventeen/pull/419

jrfnl:
@grappler Thanks for your response.

According to that reasoning:
* The wrapper around `_s_setup()` should be removed as it is hooked in.
* A wrapper should be added to `_s_categorized_blog()` as it is called directly from two different places in the theme (404 template and the `_s_entry_footer()` function)
* A wrapper should be added to `_s_infinite_scroll_render()` as it is called via another function.

grappler:
> The wrapper around `_s_setup()` should be removed as it is hooked in. 

I agree.

> A wrapper should be added to `_s_categorized_blog()` as it is called directly from two different places in the theme (404 template and the `_s_entry_footer()` function)

I am not sure if that is needed as the result is bool and the checks are very simple. I am not sure what changes would need to be made.

> A wrapper should be added to `_s_infinite_scroll_render()` as it is called via another function.

Possibly. It is also possible to use `remove_theme_support()` and redefine theme support.

jrfnl:
> > A wrapper should be added to _s_categorized_blog() as it is called directly from two different places in the theme (404 template and the _s_entry_footer() function)
>
> I am not sure if that is needed as the result is bool and the checks are very simple. I am not sure what changes would need to be made.

The `get_categories()` function provides a lot of possibilities to customize the query, so I can imagine there will be situations where people will want to change things (only categories used by a certain post type, exclude a category etc). With that in mind, I think the function being overloadable would make a lot of sense.

> > A wrapper should be added to _s_infinite_scroll_render() as it is called via another function.
>
[...]

PR #1140, opened 2017-06-28
Correct code indentation

truongwp:
The code inside if block should be indented. The current indentation makes PHPCS errors.

### DIFF
--- a/comments.php
+++ b/comments.php
@@ -27,21 +27,21 @@ if ( post_password_required() ) {
 	if ( have_comments() ) : ?>
 		<h2 class="comments-title">
 			<?php
-				$comment_count = get_comments_number();
-				if ( 1 === $comment_count ) {
-					printf(
-						/* translators: 1: title. */
-						esc_html_e( 'One thought on &ldquo;%1$s&rdquo;', '_s' ),
-						'<span>' . get_the_title() . '</span>'
-					);
-				} else {
-					printf( // WPCS: XSS OK.
-						/* translators: 1: comment count number, 2: title. */
-						esc_html( _nx( '%1$s thought on &ldquo;%2$s&rdquo;', '%1$s thoughts on &ldquo;%2$s&rdquo;', $comment_count, 'comments title', '_s' ) ),
-						number_format_i18n( $comment_count ),
-						'<span>' . get_the_title() . '</span>'
-					);
-				}
+			$comment_count = get_comments_number();
+			if ( 1 === $comment_count ) {
+				printf(
+					/* translators: 1: title. */
+					esc_html_e( 'One thought on &ldquo;%1$s&rdquo;', '_s' ),
+					'<span>' . get_the_title() . '</span>'
+				);
+			} else {
+				printf( // WPCS: XSS OK.
+					/* translators: 1: comment count number, 2: title. */
+					esc_html( _nx( '%1$s thought on &ldquo;%2$s&rdquo;', '%1$s thoughts on &ldquo;%2$s&rdquo;', $comment_count, 'comments title', '_s' ) ),
+					number_format_i18n( $comment_count ),
+					'<span>' . get_the_title() . '</span>'
+				);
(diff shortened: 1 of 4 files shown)


grappler:
Thank you for the pull request. I am surprised that there is a PHPCS error for this code as there have been no errors in travis.

This is not the only place where this type of indentation is used for pluggable functions. If we were to change it we would need to do it every where.

We are currently discussion removing the `functions_exists()` for this function https://github.com/Automattic/_s/issues/1116

I don't think the extra indentation make sense to add here.

jrfnl:
> I am surprised that there is a PHPCS error for this code as there have been no errors in travis.

That particular sniff is explicitly excluded for `_s` and I'd personally recommend excluding it for themes anyhow as for files combining PHP and HTML I've found that sniff to be quite buggy.

@truongwp You may be interested to have a look at the custom PHPCS ruleset which is included with `_s` which is a good example for a ruleset to be used for WP themes.
https://github.com/Automattic/_s/blob/master/phpcs.xml.dist

truongwp:
Sorry because I used my PHPCS rule set. But I think it still violates WordPress coding standard.
Remove checking function exists is a good idea because we can use `remove_action()`

grappler:
We we were to take the indention to the letter then we should remove the exceptions in the `phpcs.dist.xml` and make sure the indention is fix every where.

```
<exclude name="Generic.WhiteSpace.ScopeIndent.IncorrectExact" />
<exclude name="Generic.WhiteSpace.ScopeIndent.Incorrect" />
<exclude name="PEAR.Functions.FunctionCallSignature.Indent" />
```

jrfnl:
> We we were to take the indention to the letter then we should remove the exceptions in the phpcs.dist.xml and make sure the indention is fix every where.

I'd definitely wait until WPCS 0.12.0 in that case as we're working on adding some extra sniffs which should make some of the indent sniffs which are excluded slightly less buggy - at least for the fixer part -. Some sniffs work better when used together with others.
For more information, check out https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/issues/1000

grappler:
@truongwp As WPCS 0.12.0 has been released could you have a look how the correct code indentation would look like on all of the files and not just `functions.php`?

truongwp:
Hi @grappler,
I have just corrected all indentation errors when check function_exists in all files.
And when I runned phpcs, I saw that there are some other phpcs errors like: `Newline required after opening brace` or `Closing PHP tag must be on a line by itself`. Should I create a new PR to fix them?

grappler:
@truongwp No that is fine. I have a PR open for those issues. #1157

grappler:
Thanks this looks good. With the introduction of these changes we can remove
```xml
<exclude name="Generic.WhiteSpace.ScopeIndent.Incorrect" />
<exclude name="PEAR.Functions.FunctionCallSignature.Indent" />
```

To be able to remove the following exception there needs to be two fixes in `inc/custom-header.php` and `comments.php`.
`<exclude name="Generic.WhiteSpace.ScopeIndent.IncorrectExact" />`
Could you make the remaining two changes to `inc/custom-header.php` and `comments.php`?

The remaining line indent issue in `header.php` will be fixed by the changes in #1157.

truongwp:
@grappler I added the commit.

grappler:
Thanks @truongwp! I have tested this and it works great. Could you squash the commits together? Will then approve it.

truongwp:
@grappler Yes, I squashed the commits to one.

davidakennedy:
Hi @truongwp – this looks solid. Thanks for the pull request! I see one issue. It looks like support for `custom-logo` was removed, probably unintentional. 😄  Could you add that back in, and squash your commits.

truongwp:
@davidakennedy Sorry about the mistake. I restored custom-logo.

davidakennedy:
Thank you @truongwp!

PR #1142, opened 2017-06-28
Changed get_template_directory_uri function with get_theme_file_uri

vinoddalvi:


davidakennedy:
Hi @vinoddalvi! Thank you for the pull request!

In the future, please try to include some kind of description with pull requests. Even simple things. It makes it easier for maintainers to review issues and PRs in order to make decision. I'd also say that a well written issue or PR description is just as valuable as code – sometimes more so. 😄 

That said, with this feature having launched in 4.7, we should probably wait until 4.9 comes out or do some kind of check to fall back if the function isn't available. Historically, we've tried to support WordPress versions two versions back from the current one in _s.

We should also think about the `get_theme_file_path()` function as well, and look to Twenty Seventeen for the lead here. [See this post for more information](https://make.wordpress.org/core/2016/09/09/new-functions-hooks-and-behaviour-for-theme-developers-in-wordpress-4-7/).

vinoddalvi:
@davidakennedy  Thank you for your kind reply and sure i will add description in the PR henceforward.

zimnyd:
Will this change be implemented in the code?

PR #1145, opened 2017-06-29
Loop for gallery-columns should be outside gallery-item braces

VBlankSB:
The compilation gives 
.gallery-item .gallery-columns-X  { ... }
but we need
.gallery-columns-X .gallery-item { ... }

The problem doesn't exist in style.css file in the repository

### DIFF
--- a/sass/media/_galleries.scss
+++ b/sass/media/_galleries.scss
@@ -10,7 +10,7 @@
 
 	// Loops to enumerate the classes for gallery columns.
 	@for $i from 2 through 9 {
-		.gallery-columns-#{$i} {
+		.gallery-columns-#{$i} & {
 			max-width: ( 100% / $i );
 		}
 	}


grappler:
If that is the case then we can solve it by using `.gallery-columns-#{$i} & {}` instead of `.gallery-columns-#{$i} {}`.

This feature is explained here: https://css-tricks.com/the-sass-ampersand/#article-header-id-5

VBlankSB:
Fine ! Thanks ! Better way to fix the problem.

grappler:
Thank you ! This looks good.

Do you think you could squash all of the commits into one? This article explains how to do it. https://www.devroom.io/2011/07/05/git-squash-your-latests-commits-into-one/ You will need force push the changes to GitHub. You can do it with this command `git push --force`. If you would like to use another editor other than `vim` you could run `git config --global core.editor "nano"`

VBlankSB:
OK ! I read the article you advise me and I do that ! I am new user of GIT / Github, it's my first pull request ;-)
Thank you for your help and patience

grappler:
Great Valérie. Feel free to ping me on WordPress Slack if you get stuck https://wordpress.slack.com

VBlankSB:
I think it's ok now. I have also changed the description of the commit to : "use ampersand in the loop to have .gallery-item in good position".  I learned interesting things today ! Thank you Ulrich !

Issue #1149, opened 2017-06-29
Tools to build CSS from Sass


grappler:
We have previously discussed build tools for the CSS from Sass. In 2014 we decided against as we wanted to keep `_s` tool agnostic. #585, #1097, #1002, #661

> we will probably revisit once Automattic/underscores.me#31 got resolved

As it has been 2.5 years since that discussion I would like to propose a different solution.

Before merging pull requests related to CSS we need to be able to test that the Sass to CSS conversion is working properly. To be able to this properly every maintainer would need to have create a build tasks locally. Which is related to a previous discussion #615 

What I propose we document the following command which does that without needing grunt or gulp or any other build system. 

`node-sass --include-path scss sass/style.scss style.css --output-style=expanded --indent-type=tab --indent-width=1`

The only requirements for this is `node` and `npm install node-sass` which most developers are already using.

After running this command I got the following changes. I have removed some of the code style changes that have not been factored in. I think adding another check to travis for CSS code styling is another issue that we should look at in another issue. Partially mentioned in #1146 and previously discussed in #936

```diff
--- style.css
+++ style.css
@@ -49,7 +49,7 @@ Nicolas Gallagher and Jonathan Neal http://necolas.github.io/normalize.css/
 html {
        font-family: sans-serif;
        -webkit-text-size-adjust: 100%;
-       -ms-text-size-adjust:     100%;
+       -ms-text-size-adjust: 100%;
 }

 body {
@@ -426,7 +426,7 @@ input[type="submit"] {
        border-color: #ccc #ccc #bbb;
        border-radius: 3px;
        background: #e6e6e6;
-       color: rgba(0, 0, 0, .8);
+       color: rgba(0, 0, 0, 0.8);
        font-size: 12px;
        font-size: 0.75rem;
        line-height: 1;
@@ -473,10 +473,6 @@ textarea {
        padding: 3px;
 }

-select {
-       border: 1px solid #ccc;
-}
-
 input[type="text"]:focus,
 input[type="email"]:focus,
 input[type="url"]:focus,
@@ -496,6 +492,10 @@ textarea:focus {
        color: #111;
 }

+select {
+       border: 1px solid #ccc;
+}
+
 textarea {
        width: 100%;
 }
@@ -546,16 +546,6 @@ a:active {
        padding-left: 0;
 }

-.main-navigation li {
-       float: left;
-       position: relative;
-}
-
-.main-navigation a {
-       display: block;
-       text-decoration: none;
-}
-
[...]

PR #1154, opened 2017-07-01
Support RTL the right way

grappler:
@yoavf gave a talk on "[Right To Left Languages Support – The Right Way](http://wordpress.tv/2017/06/21/yoav-farhi-right-to-left-languages-support-the-right-way/)"

In his talk he discourages the use of `rtl.css` but instead use `style-rtl.css` which would be a replacement for `style.css` instead of `rtl.css` which override the ltr styles with rtl styles.

There are a couple of reasons why this is better.

- RTL users do not have worse performance by having to load an extra file for their styles.
- The RTL file can be generated automatically without having manually to change the design.

The `style-rtl.css` can be generated using a node task. After installing it `npm install -g rtlcss`, the generation happens with `rtlcss style.css style-rtl.css`. This is similar to the tool used to generate the CSS from Sass in #1149.

https://yoav.blog/2017/06/22/right-to-left-in-wordpress-the-right-way/


davidakennedy:
I'm going to leave this for now. We need to work through the automation/build process items before moving on this.

Issue #1144, opened 2017-06-28
maintainable-ize normalize


m-e-h:
In it's current form the `normalize.scss` file is a bit of a mystery. 
So is normalize being used as an un-editable drop in? It looks like it has been edited once since it was first added. 

Is it something that was planned to be kept in sync with the Normalize project? If so, what version are we currently using? There are no comments or file headers to indicate this. I'm guessing it's around version 4 since the `menu` html element is being styled. 

I say we forget about normalize (or anything else) as a drop-in and create a reset that fits and evolves with `_s`.
 
Sure, use `normalize.css` as a reference but there are others like `sanitize.css` and Bootstrap's `reboot.css` that have well documented solutions for reference too. 

If this **IS** `normalize.css`, maybe update it and keep the header info and comments in the file.
If it's **NOT** `normalize.css`, maybe rename the file to eliminate confusion. Also folks might be more willing to make pull requests on it.

davidakennedy:
Hi @m-e-h! Thanks for bringing this up.

When Normalize was originally added, it was done so to make updating a reset easier. Also, it was added so that we weren't maintaining our own bespoke reset.  See: https://github.com/Automattic/_s/commit/6c47be98c133c911c1533648225e3b720ba260d4

We may have deviated from it slightly, I'd need to search through the commit history a bit more for the reasons on that.

That said, I'd be for updating Normalize. I'd rather not add our own reset since it's more code to maintain, and Normalize seems like a battle-tested solution at this point. I also wouldn't be opposed to using another solution if it's a better fit.

m-e-h:
Normalize is good.

As for the loan modification: https://github.com/Automattic/_s/issues/895
Basically, a lot of folks were having to override one of normailize's styles. 
Which is one reason why I think maybe changing the name of the file is a good idea. A third party file like that makes things awkward. "Can we change something in the file or do we override it later in our CSS?". 
If you override, your coding sloppy. If the file's edited it's not `normalize.css` anymore or it is but it's just more difficult to update.

Maybe start with the updated `normalize.css` but change the file name so people feel comfortable suggesting changes to it. Normalize supports ie8+ which isn't totally compatible with `_s` anyway.

If we do just want to update normalize and keep it current, not having the version in the header of the file makes it very hard to maintain as a third party file. 
Is there a reason the header and comments were stripped out? Is that the preferred format if and when `normalize.scss` is updated to the current version?

grappler:
[...]

Issue #895, opened 2016-01-01
Input type search styling


samikeijonen:
All the others elements have `box-sizing: border-box;` CSS rule. But in search there is `box-sizing: content-box`:
```
input[type="search"] {
	-webkit-appearance: textfield;
	box-sizing: content-box;
}
```

This creates small styling issues on search field and I personally don't remember to remove the line every time:

```
box-sizing: content-box;
```

My question is that does most of the theme authors remove that line for better styling like I do and default themes do? And should that line be removed from `_s` theme?

sixhours:
I'd be in favor of removing this. I'm not sure why it's there, but I've also run into issues styling the search form, particularly in widgets, because it doesn't follow the rest of the site's `border-box` rules.

mariomelchor:
+1 I agree as well. I'd be in favor of removing.

corvannoorloos:
If I understand correct I'm interested to see what will happen [here](https://github.com/necolas/normalize.css/issues/486) first.

sayedwp:
+1. I have also run into the same problem. I should be removed.

KickinAss:
Me to, but i think is an error
```
//
// 1. Address `appearance` set to `searchfield` in Safari and Chrome.
// 2. Address `box-sizing` set to `border-box` in Safari and Chrome.
//

input[type="search"] {
  -webkit-appearance: textfield; // 1
  box-sizing: content-box; //2
}
```
1 and 2 do not work as the comment said

patrickhlauke:
Dupe of https://github.com/necolas/normalize.css/issues/471 ?

corvannoorloos:
> see necolas/normalize.css#471

Thanks for the update!

corvannoorloos:
With this merged I'm very much in favor of including the latest normalize.css. @samikeijonen any thoughts?

samikeijonen:
Yeah I don't see why not, but I haven't looked does `_s` have custom `normalize.css`. 

But at least `input[type="search"]` :)

corvannoorloos:
I haven't compared its current state, but it should only have its comments stripped. https://github.com/Automattic/_s/commit/6c47be98c133c911c1533648225e3b720ba260d4

karmatosed:
We have a paired down normalize, so I'd be keen we didn't merge everything in again. We also have Sass to consider. I'm ok with removing the styling though that this ticket originally was looking at.

PR #1155, opened 2017-07-01
Update Normalize.css to v7.0.0

mtomas7:
I condensed similar styles as it was in previous _s version, but if needed I have normal version too. Personally, I prefer the normal version, as it makes much easier to compare with the original Normalize.css source.

### DIFF
--- a/style.css
+++ b/style.css
@@ -44,8 +44,10 @@ Nicolas Gallagher and Jonathan Neal http://necolas.github.io/normalize.css/
 --------------------------------------------------------------*/
 
 /*--------------------------------------------------------------
-# Normalize v7.0.0
+# Normalize
 --------------------------------------------------------------*/
+
+/*! normalize.css v7.0.0 | MIT License | github.com/necolas/normalize.css */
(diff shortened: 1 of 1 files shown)


grappler:
@mtomas7 Thanks! Could you also update `sass/_normalize.scss` too?

Just for future reference the lone modification that was made before #895 has been fixed upstream so we should be ok in updating. necolas/normalize.css@fe56763

mtomas7:
Sorry, but I only worked with LESS, perhaps someone else with more experience could do this?

grappler:
From the looks of the `sass/_normalize.scss` file it is the same as the CSS so you don't need to know Sass to update it. In #1144 we were discussing about adding the version number to the Normalize header of the `style.css`.

davidakennedy:
@mtomas7 Any interest in updating things here? Thank you!

crunnells:
#1272 supercedes this PR, bringing us up to v8.0.0. Thanks for your initial work here @mtomas7!

PR #1158, opened 2017-07-03
_s: Add featured images to Post and Pages with the ability to hide them via Content Options.

thomasguillot:


grappler:
Thank you Thomas for you for the pull request. Could you add a bit of reasoning why this should be added and if there are any existing issues where this was previously discussed?

FYI The code in `inc/template-tags.php` did not pass the PHPCS checks which you can see here: https://travis-ci.org/Automattic/_s/jobs/249563298

thomasguillot:
Since the majority of themes support featured images, _s needs to implement them somehow. At WordPress.com we want to standardise their use to bring consistency for our users.
That said, not everyone might want featured images so we introduce the ability to hide them via Content Options.

inc/template-tags.php
```diff
@@ -100,3 +100,36 @@ function _s_entry_footer() {
...
+	<div class="post-thumbnail">
+		<?php the_post_thumbnail(); ?>
+	</div><!-- .post-thumbnail -->
+
+	<?php else : ?>
+
+	<a class="post-thumbnail" href="<?php the_permalink(); ?>" aria-hidden="true">
+		<?php
+			the_post_thumbnail( 'post-thumbnail', array(
+				'alt' => the_title_attribute( 'echo=0' )
```
grappler: I think `the_title_attribute( array( 'echo' => false ) )` is clearer. What is the reason for using the the_title_attribute as the alt text?

thomasguillot: It's borrowed from Twenty Sixteen.
It's to give some context to the link if the image would happen to fail.

grappler: Even though the code is GPL it still requires attribution when copying code.

The alt text should be used to the describe the image so that people with poor eyesight know what the image is. I am not convinced using the post title describes the image. https://yoast.com/image-seo-alt-tag-and-title-tag-optimization/

What I can't work out why we would have a different alt text if it is on the blog page and on a single post. I had a look at the commit history and see that the code was added by @karmatosed as part of the initial commit.

grappler:
I wrote this as a comment but is hidden as the code has changed in that section: https://github.com/Automattic/_s/pull/1158#discussion_r125899463

> Even though the code is GPL it still requires attribution when copying code.
> 
> The alt text should be used to the describe the image so that people with poor eyesight know what the image is. I am not convinced using the post title describes the image. https://yoast.com/image-seo-alt-tag-and-title-tag-optimization/
> 
> What I can't work out why we would have a different alt text if it is on the blog page and on a single post. I had a look at the commit history and see that the code was added by @karmatosed as part of the initial commit.

inc/template-tags.php
davidakennedy: @grappler It's correct as coded. It can be confusing though because of the way the image is used. In this case it's linked, which means that the alt text should be the purpose of the link – in this case, the post title. See: http://webaim.org/techniques/alttext/

grappler: Thanks, This page explained it a bit better. http://accessibility.psu.edu/images/imageshtml/

grappler:
@thomasguillot  This looks good. I would approve this. 

Do you think you could squash all of the commits into one? This article explains how to do it. https://www.devroom.io/2011/07/05/git-squash-your-latests-commits-into-one/ You will need force push the changes to GitHub. You can do it with this command `git push --force`. If you would like to use another editor other than `vim` you could run `git config --global core.editor "nano"`

I am happy to walk you through it. If not I can do it for you. You will still get your props.

grappler:
@thomasguillot Thank you for squashing the commits. Could you now rebase the branch with the `master` branch so that we can get travis to pass the tests?

samikeijonen:
`aria-hidden="true"` should be removed in link because of accessibility reasons. In general aria-hidden on focusable elements are not the best idea: the content of the node will be hidden, but a user will still be able to tab to it.

In this case user can tab into link but the feature image alt text won't be announced by screen reader because of `aria-hidden="true"`.

davidakennedy:
@samikeijonen Yeah, I forgot about that. So the best alternative here is to leave it as is, and have the double link to the same post?

samikeijonen:
Yes.

inc/template-tags.php
```diff
@@ -100,3 +100,69 @@ function _s_entry_footer() {
...
+	<div class="post-thumbnail">
+		<?php the_post_thumbnail(); ?>
+	</div><!-- .post-thumbnail -->
+
+	<?php else : ?>
+
+	<a class="post-thumbnail" href="<?php the_permalink(); ?>" aria-hidden="true">
+		<?php
+			the_post_thumbnail( 'post-thumbnail', array(
+				'alt' => the_title_attribute( 'echo=0' ),
```
grappler: I think we should use `array( 'echo' => false )`  here instead of `echo=0`.

davidakennedy:
@thomasguillot Want to refresh this based on the above feedback?

laurelfulford:
Looks like @thomasguillot made the final updates based on @grappler's review, and I've tweaked the array that was causing Travis sadness, so I'm going to merge this :)

samikeijonen:
@laurelfulford There is still `aria-hidden="true"` on tabbable item (link). It should be removed or add also `tabindex="-1"` to the link.

[Here are more details about multiple links to same page](https://make.wordpress.org/accessibility/handbook/best-practices/markup/post-excerpts-for-an-archive-template/).

laurelfulford:
Ah, thanks @samikeijonen! That was an oversight on my part -- I thought that feedback had already been addressed. 

I've made a new PR to remove the `aria-hidden="true"` here: #1243 -- just let me know if there's anything else you see that needs changing there!

PR #1161, opened 2017-07-05
Add "Next/Previous" text to single-post navigation

alaczek:
...so that users can easily know they’re looking at pagination.


grappler:
This PR brings about following change:
> Changes `Hello world!` to `Previous Hello world!` if the post title is "Hello world!".

I am not sure this is more logical. I think we could make it more logical by adding a symbol like `«`/`&#187;` and `»`/`&#187;` or by changing the design to use some type of icon. See the arrows on the both sides of https://required.com/team/ulrich/

As `single.php` is used for CPT like attachments the screenreader text would not fit. `<span class="screen-reader-text">Previous Post</span>`.

alaczek:
"Previous/Next Post" would be best, as it would give the most context, but as you pointed out, it won't work with CPTs. I'll update the screen reader text and remove the "Post".

On the other hand arrows alone don't give enough context. The "Previous" and "Next" feel like a comfortable middle ground and a good starting point.

single.php
```diff
@@ -17,7 +17,10 @@
 
 			get_template_part( 'template-parts/content', get_post_format() );
 
-			the_post_navigation();
+			the_post_navigation( array(
+				'prev_text' => '<span class="screen-reader-text">' . __( 'Previous', '_s' ) . '</span><span aria-hidden="true" class="nav-subtitle">' . __( 'Previous', '_s' ) . '</span> <span class="nav-title">%title</span>',
```
joshmcrty: Doesn't this just add the word "Previous/Next" twice? Once just for screen readers, and once visually with the `.nav-subtitle` class but hidden from screen readers? This could be simplified to just one `<span>` I think.

joshmcrty:
It seems like there should be a colon or dash or something to provide more separation/context for the word "Previous/Next" because without it the word seems to just be part of the title in the absence of additional CSS/visual styling. That way it would become "Previous: Hello world!"

Also, if arrows are a possible addition, I don't think angled quotes (« ») should be used. They _look_ like arrows, but they are quotation marks. Better to use `←` and `→` or another actual arrow character instead for semantics.

davidakennedy:
Thanks for the PR @alaczek! 😄 

Like @joshmcrty points out, we can go simpler here. I would just make the markup like this:

`<span class="nav-subtitle">' . __( 'Next', '_s' ) . '</span> <span class="nav-title">%title</span>`

I think we can just go with previous and next because it adds more context but not so much that it doesn't work with post types. I'd also prefer not to add any arrows because most of the time, a theme will style with actual icons. If they prefer not to, this way, they don't have to remove anything, and the classes allow for easy styling.

grappler:
@davidakennedy The text that i generated from the that sting is "Previous Hello world!" which for me is still not clear if "Previous" is part of the title. I would have something like "Previous: Hello world!". The title would need to added as a variable in the string as in some languages the order of the words may be different.

davidakennedy:
@grappler Right. That's a good point. We should go in that direction.

The reason that didn't stand out to me is that the need for the `:` often depends on the design. Like in Twenty Sixteen:

screen shot 2017-08-18 at 3 49 43 pm

But it makes sense to have the `:` added as a starter point.

alaczek:
Good call! I removed the screen-reader-text and added a colon, so that it reads "Previous: / Next:".

davidakennedy:
@alaczek This looks good to go. One thing I noticed though, there is an extra reference to `inc/template-tags.php:29` in the `.pot` file. Maybe an issue when you updated the file? If you could refresh and remove that, I'll merge this. Thanks! 😄 

See [this file in your PR](https://github.com/Automattic/_s/blob/142e2d503067725e4ef8448c105216be70276c14/languages/_s.pot).

single.php
```diff
@@ -17,7 +17,10 @@
 
 			get_template_part( 'template-parts/content', get_post_type() );
 
-			the_post_navigation();
+			the_post_navigation( array(
+				'prev_text' => '<span class="nav-subtitle">' . __( 'Previous:', '_s' ) . '</span> <span class="nav-title">%title</span>',
```
jrfnl: The call to `__()` should be replaced with `esc_html__()`, both in this line and the line below.

alaczek:
@davidakennedy done!

PR #1162, opened 2017-07-06
proper indent

B-07:
#### Related issue(s):

### DIFF
--- a/header.php
+++ b/header.php
@@ -8,7 +8,9 @@
  *
  * @package _s
  */
-?><!doctype html>
+
+?>
+<!doctype html>
 <html <?php language_attributes(); ?>>
 <head>
 	<meta charset="<?php bloginfo( 'charset' ); ?>">


grappler:
@B-07 Thank you for the pull request. It seems your changes have not passed the [PHPCS check](https://travis-ci.org/Automattic/_s/jobs/250675806). Could you please correct this?

Could you also fill elaborate why these changes should be made and add any issues where this ws previously discussed?

B-07:
Nit sure why it wasn't passing the test. But the thing is, Metas and Links within head should be indented, it's a common practice I think

grappler:
@B-07 Seems like you fixed it the end.  🎉 Do you think you could squash all of the commits into one? This will help with a clean history in the future. This article explains how to do it. https://www.devroom.io/2011/07/05/git-squash-your-latests-commits-into-one/ You will need force push the changes to GitHub. You can do it with this command `git push --force`. If you would like to use another editor other than `vim` you could run `git config --global core.editor "nano"`

Both [Twenty Sixteen](https://github.com/WordPress/WordPress/blob/master/wp-content/themes/twentysixteen/header.php) and [Twenty Fifteen](https://github.com/WordPress/WordPress/blob/master/wp-content/themes/twentyfifteen/header.php) have the indention.

B-07:
closed this one, creating a new pr with a proper commit

PR #1159, opened 2017-07-03
CSS: Add package.json for linting and compiling

m-e-h:
This is obviously a first pass since it hasn't really been discussed or tested beyond my Linux desktop. But I think it could help to guide and keep uniform some of the SCSS/CSS and JS commits.
A `package.json` would give us an agnostic build process. https://github.com/Automattic/_s/issues/1149
It's just `node`, which is what our CSS/JS development tools are built for.
https://github.com/WordPress-Coding-Standards/stylelint-config-wordpress
https://github.com/WordPress-Coding-Standards/eslint-config-wordpress

According to the NPM docs it serves 3 main purposes:
1. It serves as documentation for what packages your project depends on.
2. It allows you to specify the versions of a package that your project can use using semantic versioning rules.
3. Makes your build reproducible which means that its way easier to share with other developers.

This file could also, I think, take the place of and hopefully allow removal of the `.jscsrc` and `.jshintignore` files in the repo. Might be time to switch to ESLint anyway. JSCS was merged into ESLint over a year ago https://core.trac.wordpress.org/ticket/31823

Here's how it works for anybody wanting to submit a CSS pull-request:
From the _s/ root directory in the command line:
first `npm install` _or `yarn`_
then `npm run build:css` _or `yarn build:css`_
And that's it. They've linted the `style.css` for errors, compiled the SCSS, added prefixes according to the specified browser support, and automatically fixed some errors with the remaining errors printed in the terminal for manual fixing.

You can replace `build:css` with any of the other "scripts" commands listed in the file.

PR #1168, opened 2017-07-07
Format SCSS

m-e-h:
Lint and properly format SCSS so it'll produce a properly formatted `style.css`

#### Related issue(s):
https://github.com/Automattic/_s/issues/1149
https://github.com/Automattic/_s/pull/1159
https://github.com/Automattic/_s/pull/932
#893


m-e-h:
Oops. Looks like I need to squash a few commits here. Reckon I should add the compiled `style.css` too...

m-e-h:
The updated `style.css` has a couple minor quirks  due to `node-sass` and/or WP CSS standards. 
1. empty selector blocks for the menu section only appear in the SASS files.
2. single line comments on the same line as a parameter have been changed from `/* */` to `//`, so they only show in the SASS files, as well. This is because during compilation the `/* */` was being moved to it's own line.

If automating the `style.css` generation through SASS compilation the goal, I don't see many ways around making some adjustments like this. 

I did still have to make one manual fix to the `style.css`.
I think it was `node-sass` that didn't seem to like this piece of code.
```
.comment-navigation,
.posts-navigation,
.post-navigation {

	.site-main & {
		margin: 0 0 1.5em;
		overflow: hidden;
	}
}
```
It produced this in the CSS
```
.site-main .comment-navigation,
.site-main
.posts-navigation,
.site-main 
.post-navigation {
	margin: 0 0 1.5em;
	overflow: hidden;
}
```
The SCSS will probably need adjusted here but I thought that could be done after this commit?
Please let me know any thoughts on these or anything else that doesn't look right.

m-e-h:
Just noticed this has some overlap with this https://github.com/Automattic/_s/pull/1153 and this https://github.com/Automattic/_s/pull/1170

grappler:
Thank you for the PR.

As #1149 has started to question the need for Sass we will need to wait to merge this till we have a resolution.

As we are planning to update Normalize in #1155 the pull request will need to be refresh afterwards.

Ismail-elkorchi:
The formatting issues of SASS and CSS were resolved in https://github.com/Automattic/_s/pull/1391

PR #1170, opened 2017-07-09
Add a newline to end of file for each file that doesn't have one.

eflynn:

Inserts final newline to files that need it. `stylelint` barks at you if you don't have a newline at the end of a file. Also it's easier to read files on the shell with `cat`, for example. An `.editorconfig` file would enforce this, but it was debated before that it was a matter of personal choice since this is a template for WordPress themes and not an actual theme.

#### Related issue(s):

https://github.com/Automattic/_s/issues/688


grappler:
Thank you @eflynn for the PR. It seems this PR is doing the same thing as #1153.

eflynn:
Ah, thanks, I didn't see that one. Funny I tried to search for something like it and didn't find that one. Can you merge this or the other one? #1153 was posted 9 days ago...

grappler:
Yes, it seems like GitHub does not handle fuzzy search well.

As I have now mentioned in the other PR. I would like to see the new line being checked for in travis before the changes are made so that they do not get reversed by mistake in the future.

eflynn:
Git diff already warns you about this issue. I think you can add a pre-commit hook for git that would block a commit based on this issue? I'm not sure.

However, I strongly vote for the `.editorconfig` solution, because it solves the problem at the editor layer and is well supported. There is already a config for Wordpress itself:

https://github.com/WordPress/wordpress-develop/blob/master/.editorconfig

grappler:
I think it may be best to close this PR and keep the discussion to #1153 instead of two places. We can always reopen the PR if need be-

davidakennedy:
Closing this out in favor of #1153.

PR #1173, opened 2017-07-10
no-sidebar class

benoitchantre:
Taken from [Theme Components](https://github.com/Automattic/theme-components/blob/master/inc/extras.php#L27-L30)

Useful class to adjust the layout when there's no sidebar.


grappler:
If we add the extra class then we should add some styles too. e.g.
https://github.com/Automattic/theme-components/blob/master/assets/stylesheets/layout/_layout.scss#L19-L30

benoitchantre:
Thanks for pointing that @grappler, I'll update the PR to add the styles.

davidakennedy:
Thanks for the pull request @benoitchantre! This is a good change, and I agree that we should add some styles to go along with it. Pulling from Components should work nicely.

benoitchantre:
@davidakennedy The styles are now added.

grappler:
@benoitchantre Do you think you could squash all of the commits into one? This article explains how to do it. https://www.devroom.io/2011/07/05/git-squash-your-latests-commits-into-one/ You will need force push the changes to GitHub. You can do it with this command `git push --force`. If you would like to use another editor other than `vim` you could run `git config --global core.editor "nano"`

I am happy to walk you through it. If not I can do it for you. You will still get your props.

Doing a refresh should fix the failing checks.

benoitchantre:
It looks good, thanks @grappler for the link to the documentation.

davidakennedy:
Sorry to bug you again @benoitchantre! I was looking at this PR again today, and thinking through some things... 😄 

It may make more sense to add the styles in the layout stylesheets here, https://github.com/Automattic/_s/tree/master/layouts plus the Sass ones here, https://github.com/Automattic/_s/tree/master/sass/layout

Because those are the layouts where the no sidebar centering makes the most sense. Having that in the main stylesheet without a layout isn't as useful.

Let me know if this doesn't make sense. Thanks for your work on this!

benoitchantre:
No problem @davidakennedy, this is better. Thanks for having noticed that.
I have also improved a detail: instead of resetting all margins to 0, it resets only what is needed: [margin-left](https://github.com/Automattic/_s/pull/1173/files#diff-a33a89f242c8a2a05d28a85355023699R18) or [margin-right](https://github.com/Automattic/_s/pull/1173/files#diff-6e4ce17e6e9adc3085caf633521d1e5fR18).

Let me know if I need to squash the commits.

Ismail-elkorchi:
I think that it would be better to move the styles to a new file in the layouts directory, which could be named `content-no-sidebar.css` or somthing similar. The same should then be done in the Sass' layout directory.

The new Sass file should then be imported into https://github.com/Automattic/_s/blob/master/sass/site/_site.scss

IMHO this change makes more sense.

benoitchantre:
@davidakennedy: the PR has been refreshed.

sass/site/_site.scss
```diff
@@ -1,5 +1,6 @@
 // @import "../layout/content-sidebar";
 // @import "../layout/sidebar-content";
+
```
grappler: As this change is not specific to this PR, it would be best to remove it.

grappler:
I started to have another look at this. I am inclined to agree with @Ismail-elkorchi 

Repeating the same code in `_content-sidebar.scss` and `_sidebar-content.scss` is not very DRY.

Also I would change the code in SASS to something like this otherwise we are just writing CSS.
```sass
.no-sidebar {
	.content-area {
		float: none;
		margin-left: auto;
		margin-right: auto;
	}
	.site-main {
		margin-right: 0;
	}
}
```

I think we also need to mention somewhere that the styling are not automatically loaded for `.no-sidebar`. I was wondering why they were not loading when I testing the changes till I looked at the changes again.

davidakennedy:
Those are good suggestions @grappler. @benoitchantre Do you want to do another refresh based on the feedback here?

benoitchantre:
I'll squash all of the commits into one once you think its ready to merge.

sass/site/_site.scss
grappler: @benoitchantre Maybe we should add the `@import` for the new file here too?

grappler:
It looks to me if you revert https://github.com/Automattic/_s/pull/1173/commits/f3fde1a28e687402e06d6f981cedd0879bcedc47

benoitchantre:
@grappler what do you think to import `_no-sidebar.scss` at the end of `_content-sidebar.scss` and `_sidebar-content.scss`? It would keep the styles DRY and load them only if needed.

grappler:
I think it is a good idea.

benoitchantre:
@grappler, there's a conflict I cannot solve because I haven't write access to the repository.
Do you want I squash the commits now?

benoitchantre:
Thanks @grappler for having noticed that. CSS files have been fixed.

laurelfulford:
Thanks for all your work on this PR, @benoitchantre!

Issue #1175, opened 2017-07-10
Make it easier to translate underscores


Ismail-elkorchi:
HTML tags should be removed from translation strings, because they make translating very difficult for RTL languages ( like Arabic and Hebrew ).

grappler:
The problem is that in other languages like Japanese the order of sentence changes. See #1088

Ismail-elkorchi:
@grappler Thanks for your response.

What about changing this : https://github.com/Automattic/_s/blob/master/inc/template-tags.php#L69-L80

to something like this : 
```
sprintf(
  wp_kses(
    /* translators: Order of %1$s, %2$s and %3$s must be changed for languages other than Subject–verb–object ones. %1$s: 'Leave a Comment'. %2$s: 'on'. %3$s: post title */
    __( '%1$s <span class="screen-reader-text">%2$s %3$s</span>' ),
    array(
      'span' => array(
        'class' => array(),
      ),
    )
  ),	
  __( 'Leave a Comment', '_s' ),
  __( 'on', '_s' ),
  get_the_title()
)
```
This solution has the advantage of allowing us to change the order of the sentence, and in the same time, saves us from dealing with LTR and RTL in the same translation strings, when translating to RTL languages.

grappler:
As best practice a sentence should not be split up.

The problem with the suggested format is that in Japanese you would need to translate "Leave a Comment"  as "on" and "on"  as "Leave a Comment". For translators to understand this we would need to add some type of comment the translator would need to read the code to understand this.

Bidirectional text is an nightmare. This problem is not only in `_s` but in WordPress and in the World in general.

Maybe @yoavf has some tips on how best to work around it.

yoavf:
I agree with @grappler here. The fact that it's hard to translate strings with HTML in RTL is a problem with the tools that we're using - it really shouldn't be so hard. I know GlotPress isn't really good at this, so maybe we can work on improving it there.

davidakennedy:
Thanks for the issue @Ismail-elkorchi and the good discussion here. Sounds like something we can't solve in _s alone. Closing this out.

PR #1176, opened 2017-07-10
Avoide using HTML tags in translation strings

Ismail-elkorchi:
Making translation strings simpler by removing  HTML tags inside them. The translator's comments were also removed, since they are necessary only if `%s` placeholders are used.

#### Related issue(s): https://github.com/Automattic/_s/issues/1175


grappler:
Add comments to the issue #1175

PR #1178, opened 2017-07-11
Remove blank line in head

B-07:
A blank line there looks so odd (to me and in general), looks much much better without the blank line 

#### Related issue(s):n/a

### DIFF
--- a/header.php
+++ b/header.php
@@ -16,7 +16,6 @@
 	<meta charset="<?php bloginfo( 'charset' ); ?>">
 	<meta name="viewport" content="width=device-width, initial-scale=1">
 	<link rel="profile" href="http://gmpg.org/xfn/11">
-
 	<?php wp_head(); ?>
 </head>
 


jrfnl:
IMHO that blank line is there for a reason, i.e. to make it a *lot* easier to see that the code above it was HTML being added and the call below it is a function call.

WaveRoutes:
I also thought so, but twentyfifteen, sixteen don't have that blank line

jrfnl:
> but twentyfifteen, sixteen don't have that blank line

So what ? Learning from the twentysomethings for what they do right is a great thing, but let's not take over the things they do wrong.

B-07:
By no means I meant that, If there's a reason that's fine.

grappler:
I don't think we need to remove the the extra space.

davidakennedy:
I don't think we need to remove the space here – it's not a big deal. 😄 

Thank you for the PR though, @B-07!

PR #1181, opened 2017-07-13
Add theme tags

### LINKED ISSUE #1147: Adding relevant theme tags
I saw other threads regarding remove tags. But I believe we can add theme tags which are appropriate to the theme like following.

* custom-background
* custom-menu
* custom-logo
* threaded-comments
* translation-ready

ernilambar:
Adding relevant theme tags.

Fixes #1147

### DIFF
--- a/readme.txt
+++ b/readme.txt
@@ -1,7 +1,7 @@
 === _s ===
 
 Contributors: automattic
-Tags:
+Tags: custom-background, custom-logo, custom-menu, featured-images, threaded-comments, translation-ready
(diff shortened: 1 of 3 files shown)


davidakennedy:
Thanks for the pull request, @ernilambar! Looks good. Can you also add the same tags here for consistency? See: https://github.com/Automattic/_s/blob/master/readme.txt

ernilambar:
@davidakennedy theme tags added in readme file.

PR #1183, opened 2017-07-17
Fix failing travis test from PHPCompatibility changes in master

grappler:
So to prevent issues in the future with changes happening in `master` we should clone specific versions of PHPCS rulesets.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -47,11 +47,11 @@ before_script:
     - export PHPCS_DIR=/tmp/phpcs
     - export SNIFFS_DIR=/tmp/sniffs
     # Install CodeSniffer for WordPress Coding Standards checks.
-    - if [[ "$SNIFF" == "1" ]]; then git clone -b 2.9 --depth 1 https://github.com/squizlabs/PHP_CodeSniffer.git $PHPCS_DIR; fi
+    - if [[ "$SNIFF" == "1" ]]; then git clone -b 2.9.1 --depth 1 https://github.com/squizlabs/PHP_CodeSniffer.git $PHPCS_DIR; fi
     # Install WordPress Coding Standards.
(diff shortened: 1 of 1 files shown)


jrfnl:
In that case, I would suggest using Composer instead of git cloning and setting `prefer-stable=true`.

Setting the versions in stone by hard-coding specific version nrs will too easily and too quickly lead to using really outdated version.

As a side-note for these specific repositories:
* PHPCS and PHPCompatibility use `master` as the dev branch.
* For WPCS, `master` is always the latest release and should be fine to use if you want to stay up to date.

grappler:
@jrfnl Thanks, is it possible to use composer without having a `composer.json` with travis?

jrfnl:
> @jrfnl Thanks, is it possible to use composer without having a composer.json with travis?

I believe so.

> I think it would be best to wait till PHPCompatibility supports composer and switch all together.

The `master` of PHPCompatibility has been fixed now for Composer support (that's why you're seeing the break).

grappler:
I was able to do it with composer. As PHPCS is already defined as a dependency in WPCS and PHPCompatibility there was no need to require it.

The reason to install the packages as globals is so that the `vendor` directory is not added to the theme and would require excluding from any checks.

The current method is running the development version of `PHPCompatibility`. I think it would be best to use the git method till the next stable version of `PHPCompatibility` is released. If we agree on this then I will remove the last commit.

jrfnl:
> till the next stable version of PHPCompatibility is released

Expected soon ;-) That version will also be compatible with PHPCS 3.x, while still supporting PHPCS 2.x as well. The changelog will contain all the relevant information needed to fix locals installs and build scripts as we're aware that this is a breaking change.

grappler:
@davidakennedy I have remove the commit that uses composer. Once the next version of `PHPCompatibility ` is released I will create a new PR with https://github.com/grappler/_s/commit/228dfb31c18f00a6fdc377c2082aec9ea44ce0ab

PR #1184, opened 2017-07-18
Travis: Use Ubuntu Trusty by default and Precise for PHP 5.2

grappler:
Via [#41292](https://core.trac.wordpress.org/ticket/41292)/[r41072](https://core.trac.wordpress.org/changeset/41072)

Starting July 18th 2017, Travis will begin switching the default image to Ubuntu Trusty, which does not support PHP 5.2 or 5.3.

https://blog.travis-ci.com/2017-07-11-trusty-as-default-linux-is-coming

WordPress Core have just committed the above change just now via 

This is not a full fix, because Travis will be dropping precise support entirely in September which will mean we can't run tests on PHP 5.2. I will open an issue where we can track the progress in WordPress Core.


grappler:
> WordPress Core has dropped testing HHVM, I suggest the same here in a new PR

Thanks, I have planned to create a PR. Just waiting to get these few changes merged in first.

PR #1157, opened 2017-07-02
Move PHP tags on it’s own line & correct indention for WPCS 0.12.0

grappler:
This PR fixes WPCS errors which will show once `0.12.0` is released. This fixes them preemptively.

The whole list is
```
FILE: /_s/archive.php
----------------------------------------------------------------------
FOUND 3 ERRORS AFFECTING 3 LINES
----------------------------------------------------------------------
 16 | ERROR | [x] Closing PHP tag must be on a line by itself
    |       |     (Squiz.PHP.EmbeddedPhp.ContentBeforeEnd)
 27 | ERROR | [x] Newline required after opening brace
    |       |     (Squiz.ControlStructures.ControlSignature.NewlineAfterOpenBrace)
 44 | ERROR | [x] Closing PHP tag must be on a line by itself
    |       |     (Squiz.PHP.EmbeddedPhp.ContentBeforeEnd)
----------------------------------------------------------------------
PHPCBF CAN FIX THE 3 MARKED SNIFF VIOLATIONS AUTOMATICALLY
----------------------------------------------------------------------

FILE: /_s/comments.php
----------------------------------------------------------------------
FOUND 5 ERRORS AFFECTING 5 LINES
----------------------------------------------------------------------
 27 | ERROR | [x] Closing PHP tag must be on a line by itself
    |       |     (Squiz.PHP.EmbeddedPhp.ContentBeforeEnd)
 48 | ERROR | [x] Expected 1 space before closing PHP tag; 0 found
    |       |     (Squiz.PHP.EmbeddedPhp.SpacingBeforeClose)
 58 | ERROR | [x] Expected 1 space before closing PHP tag; 0 found
    |       |     (Squiz.PHP.EmbeddedPhp.SpacingBeforeClose)
 69 | ERROR | [x] Expected 1 space before closing PHP tag; 0 found
    |       |     (Squiz.PHP.EmbeddedPhp.SpacingBeforeClose)
 86 | ERROR | [x] Closing PHP tag must be on a line by itself
    |       |     (Squiz.PHP.EmbeddedPhp.ContentBeforeEnd)
----------------------------------------------------------------------
PHPCBF CAN FIX THE 5 MARKED SNIFF VIOLATIONS AUTOMATICALLY
----------------------------------------------------------------------

FILE: /_s/footer.php
----------------------------------------------------------------------
FOUND 2 ERRORS AFFECTING 2 LINES
----------------------------------------------------------------------
[...]

PR #1186, opened 2017-07-20
WPCS: Update the custom PHPCS ruleset for the release of WPCS 0.12.0

jrfnl:
* Verify that everything in the global namespace is prefixed with a theme prefix.
* Verify that no deprecated or removed WP classes are being used.
* Verify that no deprecated or removed WP function parameters are being used.

The build for this PR will not pass until a number of issues are addressed. This is being handled in a separate PR #1157

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -51,7 +51,7 @@ before_script:
     # Install CodeSniffer for WordPress Coding Standards checks.
     - if [[ "$SNIFF" == "1" ]]; then git clone -b 2.9.1 --depth 1 https://github.com/squizlabs/PHP_CodeSniffer.git $PHPCS_DIR; fi
     # Install WordPress Coding Standards.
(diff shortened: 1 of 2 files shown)


grappler:
I think we should leave out `WordPress.NamingConventions.PrefixAllGlobals` till we have fixed WordPress-Coding-Standards/WordPress-Coding-Standards#1043.

We can also remove `<exclude name="PEAR.Functions.FunctionCallSignature.Indent" />` as it does not prevent any errors anymore.

jrfnl:
> I think we should leave out WordPress.NamingConventions.PrefixAllGlobals till we have fixed

What about adding it, but commenting it out with a "Temporarily disabled until..." comment ? That way it still serves as an example for others without breaking the build for now.

grappler:
> What about adding it, but commenting it out with a "Temporarily disabled until..." comment ?

I am undecided. Part of the problem is fixed with excluding the check on `404.php`, `comments.php` and `header.php`. At the moment we do not have a way to whitelist `$content_width` and `$themecolors`.

FYI #1140 will fix the indentation errors. The PR can be updated once that is merged.

jrfnl:
I've rebased the PR as #1140 has been merged, PR #1157 should solved the remaining issues.

PR #1189, opened 2017-07-23
Add numeric pagination support

Ismail-elkorchi:
Remove the `the_posts_navigation()` in `archive.php`, `index.php` and `search.php` and replaced it with a function that uses `paginate_links()`.

Reference: https://developer.wordpress.org/reference/functions/paginate_links/

#### Related issue(s): https://github.com/Automattic/_s/pull/1161

### DIFF
--- a/inc/template-tags.php
+++ b/inc/template-tags.php
@@ -20,8 +20,8 @@ if ( ! function_exists( '_s_paging_nav' ) ) :
 		$args = wp_parse_args( $args, array(
 			'end_size' => 2,
 			'mid_size' => 2,
-			'prev_text' => __( '&#171 Previous page', '_s' ),
-			'next_text' => __( 'Next page &#187', '_s' ),
+			'prev_text' => __( '← Previous page', '_s' ),
+			'next_text' => __( 'Next page →', '_s' ),
 			'before_page_number' => '<span class="screen-reader-text">' . __( 'Page', '_s' ) . ' </span>',
 		) );
 		the_posts_pagination( $args );
--- a/inc/template-tags.php
+++ b/inc/template-tags.php
@@ -54,7 +54,20 @@ if ( ! function_exists( '_s_paging_nav' ) ) :
 		<nav class="navigation paging-navigation" role="navigation">
 			<h1 class="screen-reader-text"><?php esc_html_e( 'Posts navigation', '_s' ); ?></h1>
 			<div class="pagination loop-pagination">
-				<?php echo esc_html( $links ); ?>
+
+				<?php
+				echo wp_kses(
+					$links,
+					array(
+						'a' => array(
+							'href' => array(),
+							'class' => array(),
+						),
+						'span' => array(),
+					)
+				);
+				?>
+
 			</div><!-- .pagination -->
 		</nav><!-- .navigation -->
 		<?php
--- a/inc/template-tags.php
+++ b/inc/template-tags.php
@@ -9,70 +9,22 @@
 
 if ( ! function_exists( '_s_paging_nav' ) ) :
 	/**
-	 * Echo archive pagination in page numbers format.
+	 * Displays a paginated navigation to next/previous set of posts, when applicable.
 	 *
 	 * This is shown at the end of archives to get to another page of entries.
+	 *
+	 * @param array $args Optional. See the_posts_pagination() for available arguments.
+	 * @link https://developer.wordpress.org/reference/functions/the_posts_pagination/#parameters
 	 */
-	function _s_paging_nav() {
-		global $wp_query, $wp_rewrite;
-
-		// Don't print empty markup if there's only one page.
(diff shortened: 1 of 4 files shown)


inc/template-tags.php
```diff
@@ -7,6 +7,59 @@
...
+ /**
+  * Display navigation to next/previous set of posts when applicable.
+  */
+ function _s_paging_nav() {
+	 global $wp_query, $wp_rewrite;
+
+// Don't print empty markup if there's only one page.
+if ( $wp_query->max_num_pages < 2 ) {
+	return;
+}
```
grappler: The indention needs to be fixed here.

inc/template-tags.php
```diff
@@ -7,6 +7,59 @@
...
+		'next_text' => __( 'Next &rarr;', '_s' ),
+	) );
+
+	if ( $links ) :
+
+	?>
+	<nav class="navigation paging-navigation" role="navigation">
+		<h1 class="screen-reader-text"><?php esc_html_e( 'Posts navigation', '_s' ); ?></h1>
+		<div class="pagination loop-pagination">
+			<?php echo $links; ?>
```
grappler: This variable needs to be escaped.

Ismail-elkorchi:
@grappler Thank you for your response. These changes should be made to display pagination in page numbers format instead of Older Posts/Newer Posts format.

Numeric pagination, in contrast with Older/Newer pagination, is more user-friendly, more accessible, and more SEO friendly, and in addition, it gives feedback of the current page the website visitor is on.

Otherwise, I fixed the indention and the escaping issue in my recent commits.

The changes look like this on the front end :
_s_pagination

gatespace:
Why are you making custom functions without using `the_posts_pagination()` ?
Reference: https://developer.wordpress.org/reference/functions/the_posts_pagination/

Ismail-elkorchi:
@gatespace the custom function `_s_paging_nav()` uses a built-in wordpress function : `paginate_links()`.

The `the_posts_pagination()` function displays a paginated navigation to Older Posts/Newer Posts, while the `paginate_links()` function create paginated link in page numbers format: ( e.g.: &larr;Previous 1 … 3 4 5 6 7 … 9 Next &rarr; )

gatespace:
@Ismail-elkorchi 

`the_posts_navigation()` and `the_posts_pagination()`are different functions.
`the_posts_pagination()` use `paginate_links()`
see: `get_the_posts_pagination()` https://developer.wordpress.org/reference/functions/get_the_posts_pagination/

I attach a screenshot of the results of running `the_posts_navigation ()` and `the_posts_pagination ()`.


I think that `the_posts_pagination()` is sufficient for numeric pagination.

Ismail-elkorchi:
@gatespace 

> the_posts_navigation() and the_posts_pagination() are different functions.

Sorry, I read your response without paying attention to the fact that they were two different functions, I thought I was talking about `the_posts_navigation()`, and assumed that you were talking about it too.

Finally, I see what you meant. I updated the PR to make use of `the_posts_pagination()`.

Ismail-elkorchi:
@joshmcrty Thank you for your review. I updated the PR to take advantage of your suggestions.

grappler:
> Numeric pagination, in contrast with Older/Newer pagination, is more user-friendly, more accessible, and more SEO friendly

Are there any articles or data that shows this?

I don't think the code comments `// Previous/next post navigation.` are needed. If the code is not clear then the function name should be renamed.

Could you align the array values?

I was wondering why you choose those values but then after doing some research I realize that they are similar to that in Twenty Fifteen, Twenty Sixteen, Twenty Seventeen.

> I think you can just put the actual characters in the code: ←→

I find that is better to use the HTML code as there could be issues depending on the file ecoding https://stackoverflow.com/questions/16594636/using-html-symbol-entities-instead-of-the-actual-symbol

I was thinking about how `'<span class="screen-reader-text">' . __( 'Page', '_s' ) . ' </span>'` would be translated as it is outputted as `<span class="screen-reader-text">Page </span>1`

@mayukojpn and @waviaei How would you translate `Page 1` in Japanese? Would the order of the word and number change?

@dmtrmrv Would the translation of `Page 1 be different depending on the value in Russian?

Ismail-elkorchi:
@grappler I adjusted the code based on your feedback.

> Are there any articles or data that shows this?

See: https://yoast.com/pagination-infinite-scrolling/

dmtrmrv:
@grappler Hey!

In this case, the word _Page_ would not change depending on the page number, but the order matters. So _Page 1_ is correct. It's not the same as with comments.

davidakennedy:
Hi @Ismail-elkorchi! Thanks for the pull request here, and the work you've put into this.

Right now, I don't think there's a need to change the posts navigation. This is a tough one because that type may not work with the design of all themes. So it's better to go simpler with the existing tag and let themers make the call.

I'd like to tweak the approach in #1161 and make improvements that way. I'm going to close this out.

Issue #1190, opened 2017-07-26
Add WooCommerce Support to _s


davidakennedy:
At WordPress.com, we're beginning to build more themes with WooCommerce integration.

So it makes sense to add support much in the way support exists for Jetpack. The basics are there, but not everything, and it's easy to remove if you don't need it.

Questions around WooCommerce support have come up a few times in the past in #761 and #1057.

@tiagonoronha  and @jeffikus will start a branch and PR for this. From there, we can take feedback and evolve it, if necessary.

Some principles for this:

- It will have just the basics with some basic styling for some key components.
- We won't style on top of WooCommerce styles.
- We'll make everything easy to remove.
- We'll likely consider an addition to the generator on Underscores.me so that if you don't want it, you can just get _s without it.

The idea here is to make this as easy to use as _s and help themers learn how to do a minimal implementation for WooCommerce.

philiparthurmoore:
@davidakennedy Is this good to close now?

davidakennedy:
Yep, thanks @philiparthurmoore! Looks like this didn't autoclose.

PR #1191, opened 2017-08-01
Basic WooCommerce Integration

tiagonoronha:

Adds a basic WooCommerce integration.

#### Related issue(s):

#1190


inc/woocommerce.php
```diff
@@ -0,0 +1,239 @@
...
+if ( ! function_exists( '_s_woocommerce_cart_link_fragment' ) ) {
+	/**
+	 * Cart Fragments
+	 * Ensure cart contents update when products are added to the cart via AJAX
+	 *
+	 * @param array $fragments Fragments to refresh via AJAX.
+	 * @return array Fragments to refresh via AJAX
+	 */
+	function _s_woocommerce_cart_link_fragment( $fragments ) {
+		global $woocommerce;
```
grappler: The global here seems to be redundant.

inc/woocommerce.php
```diff
@@ -0,0 +1,239 @@
...
+		} else {
+			$class = '';
+		}
+		?>
+		<ul id="site-header-cart" class="site-header-cart">
+			<li class="<?php echo esc_attr( $class ); ?>">
+				<?php _s_woocommerce_cart_link(); ?>
+			</li>
+			<li>
+				<?php the_widget( 'WC_Widget_Cart', 'title=' ); ?>
```
grappler: According to the documentation the second argument is the instance not the arguments. https://developer.wordpress.org/reference/functions/the_widget/

It would be better to create an array of arguments instead.

inc/custom-header.php
```diff
@@ -8,6 +8,14 @@
  *
  * @link https://developer.wordpress.org/themes/functionality/custom-headers/
  *
+ * You can add the WooCommerce Mini Cart to header.php like so ...
+ *
+	<?php
+		if ( function_exists( '_s_woocommerce_header_cart' ) ) {
+			_s_woocommerce_header_cart();
+		}
+	?>
```
grappler: I don't think this code is best suited here. It requires deleting code from another area if WooCommerce is not supported.

inc/woocommerce.php
```diff
@@ -0,0 +1,239 @@
...
+ *
+ * @return void
+ */
+function _s_woocommerce_setup() {
+	add_theme_support( 'woocommerce' );
+	add_theme_support( 'wc-product-gallery-zoom' );
+	add_theme_support( 'wc-product-gallery-lightbox' );
+	add_theme_support( 'wc-product-gallery-slider' );
+}
+
```
grappler: In other places in the code we do not have the extra line break between the function and hook.

inc/woocommerce.php
```diff
@@ -0,0 +1,239 @@
...
+	 * Cart Link
+	 * Displayed a link to the cart including the number of items present and the cart total
+	 *
+	 * @return void
+	 */
+	function _s_woocommerce_cart_link() {
+		?>
+			<a class="cart-contents" href="<?php echo esc_url( wc_get_cart_url() ); ?>" title="<?php esc_attr_e( 'View your shopping cart', '_s' ); ?>">
+				<?php /* translators: number of items in the mini cart. */ ?>
+				<span class="amount"><?php echo wp_kses_data( WC()->cart->get_cart_subtotal() ); ?></span> <span class="count"><?php echo wp_kses_data( sprintf( _n( '%d item', '%d items', WC()->cart->get_cart_contents_count(), '_s' ), WC()->cart->get_cart_contents_count() ) );?></span>
```
grappler: I wonder if we can make this line a bit more readable. It is a bit long.

tiagonoronha:
@grappler thanks for the review. I've made several changes based on your comments.

cc @jeffikus

inc/woocommerce.php
```diff
@@ -0,0 +1,233 @@
...
+ *
+ * @link https://woocommerce.com/
+ *
+ * @package _s
+ */
+
+/**
+ * WooCommerce setup function.
+ *
+ * @link https://docs.woocommerce.com/document/third-party-custom-theme-compatibility/
```
tiagonoronha: 👍

inc/woocommerce.php
```diff
@@ -0,0 +1,233 @@
...
+ *
+ * Removing the default WooCommerce stylesheet and enqueing your own will
+ * protect you during WooCommerce core updates.
+ *
+ * @link https://docs.woocommerce.com/document/disable-the-default-stylesheet/
+ */
+add_filter( 'woocommerce_enqueue_styles', '__return_empty_array' );
+
+/**
+ * Add 'woocommerce-active' class to the body tag
```
tiagonoronha: 👍

inc/woocommerce.php
```diff
@@ -0,0 +1,233 @@
...
+ *
+ * @param  array $classes css classes applied to the body tag.
```
tiagonoronha: 👍

grappler:
@jeffikus The commit https://github.com/Automattic/_s/pull/1191/commits/db7f3ed04f028fbe3f4f9241643b4a6fb1c32f88 does not fix this issue I mentioned here: https://github.com/Automattic/_s/pull/1191#discussion_r131637856

One solution would be to use `wp_add_inline_style()` for loading the fonts.

jeffikus:
@grappler I understand what you are suggesting, but adding SCSS variables like how I've done in https://github.com/Automattic/_s/pull/1191/commits/db7f3ed04f028fbe3f4f9241643b4a6fb1c32f88 is consistent with the practice of setting a different plugin path within WP using a constant. 

When a developer would compile the SCSS files for WooCommerce, changing 1 path for $woocommerce__plugin-path isn't unreasonable. 

Also, adding additional PHP code for something that isn't the norm doesn't make sense in my opinion.

davidakennedy:
@tiagonoronha @jeffikus @grappler Thanks for the work here! It's looking good.

Re: https://github.com/Automattic/_s/pull/1191#discussion_r131632034 I think that's something we can change down the road, but it's not a blocker for merge.

Re: https://github.com/Automattic/_s/commit/db7f3ed04f028fbe3f4f9241643b4a6fb1c32f88 I think the variables are a nice middle ground to how it was before. I think if a developer has changed their paths, they should expect some needed tweaks in a few spots here and there. I don't think we need to go the extra mile.

Nothing else looks like a blocker to me, and we should be able to get this merged early next week.

The accompanying code for the generator is ready too, although we may adjust the wording there a bit: https://github.com/Automattic/underscores.me/pull/63

Any other concerns on this pull request?

grappler:
> When a developer would compile the SCSS files for WooCommerce, changing 1 path for $woocommerce__plugin-path isn't unreasonable.

I am not opinionated regarding Sass so the `$woocommerce__plugin-path` is ok.

> I think if a developer has changed their paths, they should expect some needed tweaks in a few spots here and there.

I don't necessarily agree.

justintadlock:
This is the final, generated CSS, correct?

```
@font-face {
  font-family: 'star';
  src: url("../../plugins/woocommerce/assets/fonts/star.eot");
  src: url("../../plugins/woocommerce/assets/fonts/star.eot?#iefix") format("embedded-opentype"), url("../../plugins/woocommerce/assets/fonts/star.woff") format("woff"), url("../../plugins/woocommerce/assets/fonts/star.ttf") format("truetype"), url("../../plugins/woocommerce/assets/fonts/star.svg#star") format("svg");
  font-weight: normal;
  font-style: normal; }
```

If theme authors submit that to the theme directory, I'd ask them to change it simply because the path wouldn't work for all users.  Essentially, I'd consider this broken. 

If you absolutely need to reference some files within a plugin, the ideal method would be to use PHP to do so.

tiagonoronha:
@grappler @justintadlock thanks for your input!

I've updated the PR with the suggested changes.

davidakennedy:
This looks good to me. I think the [one remaining item](https://github.com/Automattic/_s/pull/1191#discussion_r131632034) from the review by @grappler can be addressed down the road. Also, @jeffikus has tested this extensively.

Of course, after merge, we can always iterate. 😄

PR #1196, opened 2017-08-26
Adding a check to make sure that the ABSPATH constant is set. If it i…

taupecat:
…sn't, then it means that someone is trying to access the theme's index.php file directly (which is undesirable).

I have seen lately various attempts on sites I manage to access the theme's index.php file directly, resulting in PHP fatal errors that WordPress functions (notably, get_header()) don't exist.

By adding a check for the ABSPATH constant before get_header() on the index page, I'm avoiding these errors by quitting execution of the page. This is a precautionary measure to avoid other issues that might be exploited further on in the page's execution.

#### Related issue(s):

### DIFF
--- a/index.php
+++ b/index.php
@@ -12,6 +12,9 @@
  * @package _s
  */
 
+if ( ! defined( 'ABSPATH' ) )
+	exit();
+
 get_header(); ?>
 
 	<div id="primary" class="content-area">
--- a/index.php
+++ b/index.php
@@ -12,8 +12,10 @@
  * @package _s
  */
 
-if ( ! defined( 'ABSPATH' ) )
+if ( ! defined( 'ABSPATH' ) ) {
+
 	exit();
+}
 
 get_header(); ?>
 


taupecat:
Hi, just wondering if anyone had had a chance to review this yet? Thanks.

philiparthurmoore:
@taupecat When it's reviewed you'll either get a ping or see a label or something on it. A8C will be gone for another week and I'm playing minor cleanup with @grappler. Cheers.

grappler:
I am not keen to add this as we would need to add this every php and not just the `index.php`.

It is better to solve the issue on the server. For an apache server you can do so by adding the following snippet to the `.htaccess` in the `wp-content` folder.

```
<Files *.php>
deny from all
</Files>
```

If you are using nginx then this post should help. https://www.bjornjohansen.no/block-access-to-php-files-with-nginx

You can see how this is working for me: https://wpbern.ch/wp-content/themes/twentyseventeen/index.php

philiparthurmoore:
Agree completely with @grappler here. Seems like a server-side issue and not something that should litter _s.

grappler:
@taupecat I am closing the issue for now but we can discuss it further if need be.

PR #1197, opened 2017-08-30
Fixed hard-coded URL issue given by Theme Check plugin

rupokify:
The _footer.php_ file had hard-coded URL which would generate the following issue in [Theme Check](https://wordpress.org/plugins/theme-check/) plugin:

_INFO: Possible hard-coded links were found..._

Updated that hard-coded URL so users won't need to fix it by themselves every time they use this.

#### Related issue(s):
Didn't find any related issue. Instead of creating an issue, I'm sending PR with the fix.

### DIFF
--- a/footer.php
+++ b/footer.php
@@ -1,4 +1,4 @@
-<?php
+'<?php
 /**
  * The template for displaying the footer
  *
@@ -22,7 +22,7 @@
 			<span class="sep"> | </span>
 			<?php
 				/* translators: 1: Theme name, 2: Theme author. */
-				printf( esc_html__( 'Theme: %1$s by %2$s.', '_s' ), '_s', '<a href="' . esc_url( __( "https://automattic.com/", "_s" ) ) . '">Automattic</a>' );
+				printf( esc_html__( 'Theme: %1$s by %2$s.', '_s' ), '_s', '<a href="' . esc_url( __( 'https://automattic.com/', '_s' ) ) . '">Automattic</a>' );
 			?>
 		</div><!-- .site-info -->
 	</footer><!-- #colophon -->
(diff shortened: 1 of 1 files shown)


grappler:
Thank you for the pull request.

> INFO: Possible hard-coded links were found...

As you can see that is an info and not a warning or error. The reason that this info notice is given so that the reviewer can check what the link is and then decide if it is malicious or spam. The link is question is neither so it is OK.

Making the string translatable does not does not make it more useful as Automattic does not have any translation of their site.

For these reasons I am going to close the PR. We can still continue to discuss it after the Pull Request is close.

rupokify:
Thanks for the explanation. I really appreciate it.

But before the theme goes to submission, a theme developer also checks his theme with Theme Check plugin and as this INFO is shown in the "One or more errors were found for XXXXX" section, it takes a handful amount of time for a new developer to find out that it should not block his theme for getting approved. Plus, one less line in the Errors section. So I thought, it was a good idea to escape that URL as many people might mistook it as _s is giving code which generates errors.

As you said, automattic.com is neither a malicious nor a spam link, and we all know that, I still think we can ship the URL escaped. What do you think?

I agree with the translatability. So I sent another commit removing that.

Let me know your thoughts on this. Cheers.

grappler:
> As you said, automattic.com is neither a malicious nor a spam link, and we all know that, I still think we can ship the URL escaped. What do you think?

There is no need to escape a hardcoded url as when reviewing the code you can see what the data is. One only needs to escape data that is coming from another source.

It would be better to improve the wording of the notice or not showing these info notices to theme authors when they are uploading a theme. The best place to discuss this would be https://meta.trac.wordpress.org/report

PR #1204, opened 2017-09-13
Use h1 in site-title if either front or posts page instead of needing to have both

ronilaukkarinen:

Changed if sentence to OR instead of AND. This is my suggestion to have more failsafe h1 on front page.

#### Related issue(s):

Noticed that h1 is not printed anywhere if the site has front page as a regular page and posts page on different sub page and not on the front page. I think this is a mistake logically. Prove me wrong and don't accept this PR :) btw, this is my first PR to _s, hello!


samikeijonen:
Hi @ronilaukkarinen, I think I have seen you somewhere :)

I think there are couple of reasons why there is AND.

1. In `index.php` there is already `<h1>` like this:

```php
if ( is_home() && ! is_front_page() ) : ?>
	<header>
		<h1 class="page-title screen-reader-text"><?php single_post_title(); ?></h1>
	</header>

<?php
endif;
```

Using OR in site title code would trigger two `<h1>`, not end of the world though.

2. If the site has front page as a regular page, that page title is used as `<h1>`. Same thing with OR in site title would trigger two `<h1>` in front page, not end of the world though.

With that said I'm pretty sure `_s` devs have made a decision that there will be only one `<h1>` in every view.

ronilaukkarinen:
Hello sir @samikeijonen, yes, you look familiar too ;) haha. 

Yes, I agree, there should be only one h1 in every view. Didn't realize the index.php if sentence, for that I'm sorry. 

I guess I see the "conflict" here, but I'm still confused. And that is, because underscores doesn't use `front-page.php`, but many people do (like me!). What do you propopse to do in my scenario, if I'd use "the _s way", without changing much header.php? In my case I should just remove that h1 from index.php to prevent having double. But as for the generic solution?

samikeijonen:
The `<h1>` behaviour should be the same even if you use `front-page.php`. If the site has front page as a regular page, that page title is used as `<h1>`. We can continue discussing in our language locally:)

grappler:
If I understand correctly you are using `front-page.php` and you would like the site name to be a `<h1>`.

The solution will be different depending if you are doing this on a client site or for a publically released theme.

When you add `front-page.php` the sites reading settings stop working. This is normally not a big issue on a clients site but for publically released themes it is. You can add a custom logic to the `front-page.php` so that the reading settings continue to work.

Once you add a custom logic you will need change how the `h1` is displayed. In your case it may be best to change the check to  `if ( is_front_page() ) : ?>`.

The reason we have `if ( is_front_page() && is_home() ) : ?>` is so that the `h1` is only display if the front page is a blog. If it was a static page the then the page title will show as a `h1` and on the blog page there may be a "Blog" title that is `h1`.

ronilaukkarinen:
@grappler Thank you for the added explanation. In my case there is no page title on the front page, only slogan, which is not necessarily best for h1 when considering SEO. That's what I did in this case indeed, removed the is_home. I see the logic now, makes sense and serves more people this way. 👍

PR #1205, opened 2017-09-13
Travis: build against highest available PHP version

jrfnl:

The new Trusty images as per Sept 7 include an image for PHP 7.2 (even though it hasn't been released yet).

The `_s` build tests are run against the lowest/highest supported PHP version for 5 and 7.
As `7.2` is now available, it should replace the `7.1` build.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,8 +29,8 @@ matrix:
           env: SNIFF=1
         # aliased to a recent 7.0.x version
         - php: '7.0'
-        # aliased to a recent 7.1.x version
-        - php: '7.1'
+        # aliased to a recent 7.2.x version
+        - php: '7.2'
         # aliased to a recent hhvm version
         - php: 'hhvm'
 


philiparthurmoore:
@jrfnl Thanks for this. To be clear, are you keeping `7.0` here and `7.2`? Or only bumping to `7.2`? My mind isn't quite what it used to be, so I want to make 100% sure what your intentions are here. Cheers.

jrfnl:
> To be clear, are you keeping 7.0 here and 7.2? Or only bumping to 7.2? My mind isn't quite what it used to be, so I want to make 100% sure what your intentions are here.

@philiparthurmoore The `_s` build script was testing against the highest and the lowest supported versions for PHP 5 and 7.
These were previously:
* PHP 5.2 and 5.6
* PHP 7.0 and 7.1

The "highest" PHP 7 version has now gone up from 7.1 to 7.2 for Travis. Nothing else has changed.

Issue #1202, opened 2017-09-11
Bringing back ARIA roles?


samikeijonen:
We removed ARIA roles in #1132. But WAI-ARIA 1.1 and HTML 5.1 specifications have been updated. This is the biggest change for ARIA roles:

<blockquote>&lt;header&gt; and &lt;footer&gt; elements will now only be exposed as header and footer, and banner and ContentInfo landmarks respectively, if these elements are direct descendants of the body tag, and therefore are scoped for the whole page.</blockquote>

In `_s` header and footer are not direct descendants of the body tag, they are inside `#page` wrapper.

My question is, should we bring back `<header role="banner">` and `<footer role="contentinfo">`?

Here is [more information about the ARIA roles subject](https://www.marcozehe.de/2017/08/30/firefox-57-will-less-chatty-screen-readers-situations/).

samikeijonen:
It doesn't mention `<main>` but let's talk about that too.

davidakennedy:
My feeling on this is: Yes. Kind of frustrating since we just removed them. 😄 

We could also remove the wrapper container, but I have a felling that's useful in a lot of cases for designers and developers. Curious to hear what others think.

joedolson:
Should put header and footer back; there shouldn't be any need to do that with main. 

I think that removing the wrapper container doesn't do anything to solve the problem - as soon as an author added one back in, the roles would be broken again.

philiparthurmoore:
FWIW I think the roles should be added back in, for precisely the reason that @joedolson mentioned. We cannot control how people use _s, but we can control how it's distributed, and doing it in a way that is developer-proof, as far as roles go, seems sensible.

asuh:
From an outside voice and perspective, I vote that `#page` should be dropped instead of adding back aria roles to header and footer elements.

As of October 2017, after Windows 10 Fall Creators Update is released, all modern browsers support CSS Grid Layout. Together with mobile-first responsive design and CSS Grid Layout, the need for a global wrapper element is more of a past relic. Pages used to need global auto margins and fixed widths to be centered and this was good when we weren't really considering devices smaller than a laptop/desktop.

[...]

PR #1212, opened 2017-10-02
Re-add `banner` and `contentinfo` roles to header and footer elements.

pattonwebz:
Re-add some ARIA roles back to header and footer elements.

#1202

### DIFF
--- a/footer.php
+++ b/footer.php
@@ -13,7 +13,7 @@
 
 	</div><!-- #content -->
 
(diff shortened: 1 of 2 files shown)


Ismail-elkorchi:
Thanks @pattonwebz for the PR. I am going to close this for now as it is not an issue anymore. See https://github.com/Automattic/_s/issues/1202#issuecomment-615699498.

PR #1216, opened 2017-10-13
Prevent direct access to theme files

dashkevych:
Recently, I've noticed that my log file contains a lot of PHP errors that display this message **"PHP Fatal error: Uncaught Error: Call to undefined function get_header()..."**. I've checked the file that was mentioned in the error message and the *get_header()* function was there. I was a little bit confused because the *get_header()* function is a part of the core and I have never modified WordPress core files.

I run few tests and I've found the reason why I get these messages. It looks like somebody was trying to directly access theme files. For example, if I open this URL on my local machine (I am using VVV): 

> http://src.wordpress-develop.dev/wp-content/themes/_s/index.php 

then I get this error message:

> Fatal error: Uncaught Error: Call to undefined function get_header() in /srv/www/wordpress-develop/public_html/src/wp-content/themes/_s/index.php:15 Stack trace: #0 {main} thrown in /srv/www/wordpress-develop/public_html/src/wp-content/themes/_s/index.php on line 15

When I access files that are located in the **template-parts** folder through the URL, I do not see any error message on the screen but I see it in the log file. For example, by using this URL:

> http://src.wordpress-develop.dev/wp-content/themes/_s/template-parts/content.php

I get this error message:

> [13-Oct-2017 04:12:59 UTC] PHP Fatal error:  Uncaught Error: Call to undefined function the_ID() in /srv/www/wordpress-develop/public_html/src/wp-content/themes/_s/template-parts/content.php:12
Stack trace:
#0 {main}
thrown in /srv/www/wordpress-develop/public_html/src/wp-content/themes/_s/template-parts/content.php on line 12

By preventing direct access to theme files, I was able to resolve the issue.

### DIFF
--- a/404.php
+++ b/404.php
@@ -7,6 +7,11 @@
  * @package _s
  */
 
+// Exit if accessed directly.
+if ( ! defined( 'ABSPATH' ) ) {
+	exit;
+}
+
 get_header(); ?>
 
 	<div id="primary" class="content-area">
--- a/archive.php
+++ b/archive.php
@@ -7,6 +7,11 @@
  * @package _s
  */
 
+// Exit if accessed directly.
+if ( ! defined( 'ABSPATH' ) ) {
+	exit;
+}
+
 get_header(); ?>
 
 	<div id="primary" class="content-area">
--- a/comments.php
+++ b/comments.php
@@ -10,6 +10,11 @@
  * @package _s
  */
 
+// Exit if accessed directly.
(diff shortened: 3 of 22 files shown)


grappler:
Thank you for the pull request. We recently had a similar pull request #1196 where we decided that this is something that should be fixed on the server and not the theme.

dashkevych:
Oops, I missed this pull request :) I agree with you that `.htaccess` solution is better. Thank you for the info! :)

PR #1220, opened 2017-10-25
Remove unnecessary !

B-07:
#### Related issue(s):

### DIFF
--- a/style.css
+++ b/style.css
@@ -1,4 +1,4 @@
-/*!
+/*
 Theme Name: _s
 Theme URI: https://underscores.me/
 Author: Automattic


davidakennedy:
@B-07 Thanks for the PR. Great catch!

One thing I'll mention: please put an explanation of your changes in the PR. No matter how simple. It's tedious sometimes, yes. However, this is a good habit to learn as a developer because:

- You get better at explaining things.
- You make a project maintainer's life easier.
- It's helpful when digging back through history, wondering why a certain change was made.
- We kindly ask you to do it by providing a PR template. 😄 

Thanks again for your contributions!

B-07:
@davidakennedy Thank for the suggestions, I'll follow those from now on for sure.

grappler:
@davidakennedy I don't think this was supposed to be removed. We added it deliberately in https://github.com/Automattic/_s/commit/e9c7d73be804365e44934f2b01b70e0110bcef13 in #770

davidakennedy:
Ah, I looked and missed that @grappler! Thanks for catching it. Will revert.

PR #1222, opened 2017-10-25
Update year to the latest

B-07:
#### Related issue(s): Update the year to 2017. I'll set a reminder to change it in 2 months :)


davidakennedy:
@B-07 We're always slow on this. 😄 

Needs to be updated here too: https://github.com/Automattic/_s/blob/master/sass/style.scss

Update your PR when you have a chance. Thanks!

B-07:
My bad, I forgot about scss. Closed it and continued with #1223

PR #1228, opened 2017-11-01
Make comments number comparison work as expected

SergeyBiryukov:
`get_comments_number()` returns a numeric string, not an integer, so the strict check for comments number in `comments.php` doesn't work as expected and always returns false.

See https://core.trac.wordpress.org/changeset/41729 for the same issue in Twenty Sixteen.

### DIFF
--- a/comments.php
+++ b/comments.php
@@ -28,7 +28,7 @@ if ( post_password_required() ) {
 		<h2 class="comments-title">
 			<?php
 			$comment_count = get_comments_number();
-			if ( 1 === $comment_count ) {
+			if ( '1' === $comment_count ) {
 				printf(
 					/* translators: 1: title. */
 					esc_html_e( 'One thought on &ldquo;%1$s&rdquo;', '_s' ),


awps:
Instead of doing this 
```
if ( '1' === $comment_count ) {
```
you should convert the variable to integer and then check it:
```
if ( 1 === absint( $comment_count ) ) {
```

If you need to check for a non-negative number, always use `absint`.

Ismail-elkorchi:
We can use simply `intval()`. `absint()` use `intval()` wrapped around `abs()` which is unnecessary in this case.

See: `absint()`, `intval()` and `abs()`
https://developer.wordpress.org/reference/functions/absint/
http://php.net/manual/en/function.intval.php
http://php.net/manual/en/function.abs.php

awps:
@Ismail-elkorchi it really doesn't matter which one we use. The idea here is to do a strict comparison of two numbers not of two numeric strings.

jrfnl:
@awps @Ismail-elkorchi Actually it does matter. A numeric string `1.5` would convert to integer `1`, but `1-b my lane` would also convert to integer `1` , so doing a strict string based comparison when the return value of the function call is expected to be a string is the better & safer solution.

jrfnl:
Just wondering what's the hold up for merging this fix ?

laurelfulford:
Thanks @SergeyBiryukov!

PR #1229, opened 2017-11-01
Clarify "singularly identifiable articles"

SergeyBiryukov:
`_s_pingback_header()` description refers to "singularly identifiable articles".

"Article" is not a native WordPress term, so I think it would be helpful to clarify what this means: single posts, pages, or attachments.


crunnells:
@SergeyBiryukov Looks good to me! Thanks for your work here.

Issue #1230, opened 2017-11-05
Wrong localization function within `printf()`


webmandesign:
Make sure not to use echoing localization functions inside `printf()` and `sprintf()` functions.

webmandesign:
Fixed with #1231

PR #1231, opened 2017-11-05
Fixing localization function inside `printf()`

webmandesign:

Making sure we don't use echoing localization functions within `printf()` and `sprintf()`.

#### Related issue(s):

Fixing #1230


jrfnl:
Just wondering what's the hold up for merging this fix ?

laurelfulford:
Thanks for the fix, @webmandesign!

Issue #1215, opened 2017-10-12
Re-organize SASS structure


samikeijonen:
While talking with several developers I sometimes get feedback that `sass` folder structure is complicated and hard to manage. I tend to agree.

I investigate this more and suggest small tweaks that we might do.

dingo-d:
It is a bit all over the place. This is the folder structure that we use at the firm I work: 

https://github.com/infinum/wp-boilerplate/tree/master/wp-content/themes/theme_name/skin/assets/styles

There is a clear separation of util functions and components, as well as of admin side.

samikeijonen:
@dingo-d At first look I like it:)

justnorris:
I made "unstyled" starter theme based on _s and while migrating Sass to Stylus I also refactored the folder structure quite a bit too.

https://github.com/Themetally/unstyled/tree/master/assets/style


Stylus is able to easily take advantage on Globbing, so the main file can look quite simple like this:
```
// Stylus Configuration & Mixins
@require "_config/*"
@require "_mixins/*"

/*--------------------------------------------------------------
 * Base
 *--------------------------------------------------------------*/
@require "base/reset/*"
@require "base/*"
@require "base/**/*"

/*--------------------------------------------------------------
 * Modules
 *--------------------------------------------------------------*/
@require "modules/*"

/*--------------------------------------------------------------
 * Site
 *--------------------------------------------------------------*/
@require "site/*"
@require "site/**/*"
```

I like the general idea behind `_s` folder structure, but the root level has too many folders. My solution was to group everything in 3 parts:

0. `/config/` and `/mixins/` - before anything gets styled, config files are loaded
1. `/base/` - the resets, general element styles, nothing with class-level selectors. Only plain elements here.
2. `/modules/` - the common WordPress element styles/resets, like Gallery, Captions, Screen-readers, etc.
3. `/site/` - Everything else - this is where the "the real styling" starts

Everything further down in a folder is self-explanatory. This has worked for me personally quite well.

samikeijonen:
> I like the general idea behind _s folder structure, but the root level has too many folders.

I agree that there might be too many folders, and that comes up in conversations. 

Your folder is easy to read, I can tell you that:)

davidakennedy:
[...]

PR #1232, opened 2017-11-05
Structure SASS folder

samikeijonen:
## Changes proposed in this Pull Request

- Simplify SASS folder structure. Conversation is happening in the issue #1215.
- The main goal is to have more readable structure.
- SASS folder structure should be easily maintained.
- Structure should be inline with current best practises.
- Structure should be good starting point in the bigger projects also.

## Related issue(s)

#1215


bahiirwa:
+1. This is simple to use. Thanks for the work input.

PR #1237, opened 2017-11-27
Prevent variables comments showing in the production file

satrya:
The variables comments do not need to show up in the production file.


grappler:
Thank you @satrya 

Do you think it would make sense to change the other titles too in `woocommerce.scss` so that the titles are not duplicated in `woocommerce.css`?

samikeijonen:
Let's hold this PR for a while. We have lot's of SASS/CSS related tickets coming up which we need to look first.

Ismail-elkorchi:
Closing as this was fixed in https://github.com/Automattic/_s/pull/1391.

PR #1238, opened 2017-12-04
Showing search input field default close button.

maheshwaghmare:
The search input field default close button was hidden though CSS property `-webkit-appearance: none;`.
Search Close Button

#### Related issue(s):

### DIFF
--- a/sass/_normalize.scss
+++ b/sass/_normalize.scss
@@ -175,11 +175,6 @@ input[type="number"]::-webkit-outer-spin-button {
 	height: auto;
 }
 
-input[type="search"]::-webkit-search-cancel-button,
-input[type="search"]::-webkit-search-decoration {
-	-webkit-appearance: none;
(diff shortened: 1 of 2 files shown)


grappler:
The CSS is part of Normalize. That is why I don't think we should remove it: https://github.com/necolas/normalize.css/blob/master/normalize.css#L383-L386

m-e-h:
I'm not a huge fan of this either but I guess it does "normalize" browser behavior. Hopefully normalize will revisit this https://github.com/necolas/normalize.css/issues/685 in it's next release.

grappler:
As this is already documented in Normalize I will close this PR as the solution would be to update Normalize once it is released instead of removing the CSS.

@maheshwaghmare If you want to create an issue so that we can track this that would be great.

maheshwaghmare:
Okay.

PR #1243, opened 2017-12-14
Remove aria-hidden='true' from a tabbable item in the featured image

laurelfulford:
This PR removes the `aria-hidden="true"` from the link in the featured image markup to improve accessibility.

See #1158 for more information.

### DIFF
--- a/inc/template-tags.php
+++ b/inc/template-tags.php
@@ -122,7 +122,7 @@ function _s_post_thumbnail() {
 
 	<?php else : ?>
 
-	<a class="post-thumbnail" href="<?php the_permalink(); ?>" aria-hidden="true">
+	<a class="post-thumbnail" href="<?php the_permalink(); ?>">
 		<?php
 			the_post_thumbnail( 'post-thumbnail', array(
 				'alt' => the_title_attribute( array(


rianrietveld:
Please keep the  `aria-hidden="true"` and add t`abindex=-1`
For the reason why see:
https://make.wordpress.org/accessibility/handbook/best-practices/markup/post-excerpts-for-an-archive-template/

pattonwebz:
I too am in favour reverting this so that `aria-hidden="true" comes back and `tabindex=-1` can be added in.

If it would be preferable to keep this merged and create a septate PR to add those I can provide that PR, just let me know :)

postphotos:
Echoing @pattonwebz, the more accessible tweak here is to do as @rianrietveld has suggested.

We recommend either revising above, or closing this PR #1243 and starting another request with a better accessibility method. Let us know if you'd like one of us to create a pull request to help expedite this.

rianrietveld:
@laurelfulford sorry for this, I should have asked @samikeijonen first, as he's may more on top of things here and and already convinced you :-)
.

laurelfulford:
Sorry I missed the feedback here! I'm happy to defer to you all on this -- it does sounds like `tabindex=-1` would be a better choice. Looking back at the comment in #1158, I think I misunderstood the requested in the first place. 

I'll close this PR and create a fresh one with the recommended fix. 

Thanks @rianrietveld @pattonwebz & @postphotos!

PR #1246, opened 2017-12-19
Use CSS grid for gallery styles

gooklani:
Updated gallery item styles to generate a block-level grid

### DIFF
--- a/sass/media/_galleries.scss
+++ b/sass/media/_galleries.scss
@@ -1,19 +1,20 @@
 .gallery {
 	margin-bottom: 1.5em;
+	display: grid;
+	grid-gap: 1.5em;
 }
 
 .gallery-item {
 	display: inline-block;
 	text-align: center;
-	vertical-align: top;
 	width: 100%;
+}
 
-	// Loops to enumerate the classes for gallery columns.
-	@for $i from 2 through 9 {
+// Loops to enumerate the classes for gallery columns.
+@for $i from 2 through 9 {
 
-		.gallery-columns-#{$i} & {
-			max-width: map-get($columns, $i);
-		}
+	.gallery-columns-#{$i} {
+		grid-template-columns: repeat($i, 1fr);
 	}
 }
 
--- a/style.css
+++ b/style.css
@@ -905,46 +905,47 @@ object {
 ## Galleries
 --------------------------------------------------------------*/
 .gallery {
-	margin-bottom: 1.5em;
+    margin-bottom: 1.5em;
+	display: grid;
+	grid-gap: 25px;
 }
 
 .gallery-item {
-	display: inline-block;
-	text-align: center;
-	vertical-align: top;
-	width: 100%;
+    display: inline-block;
+    text-align: center;
+    width: 100%;
 }
 
-.gallery-columns-2 .gallery-item {
-	max-width: 50%;
+.gallery.gallery-columns-2 {
+    grid-template-columns: repeat(2,1fr);
 }
 
-.gallery-columns-3 .gallery-item {
-	max-width: 33.33%;
+.gallery.gallery-columns-3 {
+    grid-template-columns: repeat(3,1fr);
 }
 
-.gallery-columns-4 .gallery-item {
-	max-width: 25%;
+.gallery-columns-4 {
+    grid-template-columns: repeat(4,1fr);
 }
 
-.gallery-columns-5 .gallery-item {
-	max-width: 20%;
+.gallery-columns-5 {
+    grid-template-columns: repeat(5,1fr);
 }
 
-.gallery-columns-6 .gallery-item {
-	max-width: 16.66%;
+.gallery-columns-6 {
+    grid-template-columns: repeat(6,1fr);
 }
 
-.gallery-columns-7 .gallery-item {
-	max-width: 14.28%;
+.gallery-columns-7 {
+    grid-template-columns: repeat(7,1fr);
 }
 
-.gallery-columns-8 .gallery-item {
-	max-width: 12.5%;
+.gallery-columns-8 {
+    grid-template-columns: repeat(8,1fr);
 }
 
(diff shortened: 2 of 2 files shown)


m-e-h:
Unfortunately, for now, `_s` still needs to support IE11. :disappointed:

grappler:
Thank you @gooklani for the pull request.

From the code styling perspective, we use tabs and not spaces so that would need to be changed.

As we currently have CSS and Sass files, both would need to be changed.

Could you mention what browser support is needed for this feature so that we can discuss it?

A screenshot of the before and after would be helpful to see any changes.

Thank you!

bahiirwa:
For browser support, could the @supports (display: grid) {} rule help? Or use css hierarchy by implementing css grid before the older css rules to allow overrides.

gooklani:
* Chrome - Full support as of March 8, 2017 (version 57)
* Firefox - Full support as of March 6, 2017 (version 52)
* Safari - Full support as of March 26, 2017 (version 10.1)
* Edge - Full support as of October 16, 2017 (version 16)
* IE11 - No support for current spec; supports obsolete version

Internet Explorer 11 will never be updated to the new specification, and this is why grid layouts are not supported in IE.

Here's the complete picture:
https://caniuse.com/#feat=css-grid

@bahiirwa good idea, the @supports CSS at-rule enables us to specify declarations which depend on a browser's support for one or more specific CSS features.


@grappler What do you think? Would you like me to implement this so far?
https://codepen.io/primalivet/pen/ryjKmV

grappler:
I think it is good to use modern CSS features. As IE still has 3% market share we need to make sure there is a backwards compatible solution. Also there are a few mobile browser that do not support it also.

It would be good to get more feedback from Frontend Developers.

@mor10 has been pushing CSS grid. It would be interesting to hear his thoughts.

mor10:
Grid is the right layout module here. If a fallback is necessary, the best option is to provide non-grid rules as the default and wrap the grid rules an an @supports rule.

However, there's a philosophical question that needs to be addressed here which has implications beyond the gallery: Should grid be used as the primary layout module for _s as a whole? If it's used in the gallery, it should be used everywhere which means _s needs to be refactored substantially. This in turn begs the question of why _s should not use grid as it's primary layout module. The immediate answer will be backwards compatibility, but is this a good reason? Is it _s' job to delay implementation of modern best practices for the sake of backwards compatibility? My personal opinion is no. _s should provide a modern baseline for development and the individual theme developer who uses it has the responsibility of providing whatever backwards compatibility is required for their project.

m-e-h:
:100: for refactoring _s layout as a whole. Floats have needed needed to go for a while, IMO. https://github.com/Automattic/_s/pull/682
And grid is surely the way to go for that as it makes short work out of complex layouts. But for something as simple as this basic gallery where the items are equal in size, what's the benefit of grid over the current inline-block implementation or even flexbox? I've not used grid a ton but it appears both the behavior and the amount of code would be equal with this PR. Is that correct?

bahiirwa:
@mor10 that personal opinion makes sense. It defines a starter theme. Not completing the job for developer.

"implementation of modern best practices" should be key and where possible we should do backward compatibility.

LittleBigThing:
Great idea to use CSS grid for the gallery and maybe for the whole _s theme.
I have 2 possible contributions on the discussion:
1. How would using Grid affect Jetpack’s gallery styling? If it would at all. And other popular plugins interacting with the gallery.
2. It seems to me that implementing grid for the whole _s theme instead of just the galleries could also be interesting since there is not a lot of layout stuff except for the floats for the sidebars and the main menu. We might work with optional SASS partials that can be commented out: a partial using floats if backwards compatibility is needed and one with Grid for more modern/adventerous solution. Even something similar using Flexbox for the navigation menu.

gooklani:
CSS Grid solves the layout structure problem by allowing designers to control both columns and rows simultaneously and in addition to that, you would be able to use Grid Gap to manage the spacing between the elements instead of tricky implementation of margin or padding.

@m-e-h I'm not sure why you are strictly taking sides against this grid implementation but to me being on the cutting edge of development is something we should want to do from anywhere or anyplace we could.

I assume this could be a start of something new and we can extend it to cover all aspects of the theme layout eventually.

bahiirwa:
On hind sight, there were issues to do with implementing framework like bootstrap etc in _s. Common consensus was leave that to developer. Doesn't using grid in the whole of _s going to brig issues with a developer choosing to use a CSS framework?

mor10:
@m-e-h like @gooklani says, using grid for the gallery opens the door for more advanced layouts, and more responsive ones too. For example, it would not be hard to make a gallery layout where some of the images are twice the size of others, and even make this dependent on how many images are in the gallery. It also makes it easier to style the gallery as a whole (because it wraps and clears properly) and gallery items individually.

mor10:
I have a fully refactored version of _s set up for grid in a private repo already. I can clean it up and submit it as a pull requests next week so you can all take a look. A lot is changed including nesting, clearings, and other features.

mor10:
@bahiirwa CSS frameworks like Bootstrap and Foundation were introduced to provide a solution to the lack of proper grid layout tools on web platform. Now that CSS Grid is live, the framework approach of complex HTML-based grids is no longer recommended, and actively works against the way the web works. Some frameworks will adopt grid as their layout module. Others will stick with old methods and will hold bot themselves and the web back as a result. In my opinion it would be counterproductive to avoid grid to preserve support for outdated methods.

samikeijonen:
I'm all in for using Grid everywhere even if it's pretty basic stuff in `_s`. Also noting that I'm 100% sure `_s` is going to be "Gutenberg ready" in 2-4 months.

Ismail-elkorchi:
Hi @gooklani  👋 ! Would you be still interested in updating this PR now that https://github.com/Automattic/_s/pull/1251 was merged ? If not, no worries, I can pick up the work where you left off if you want, just let me know.

PR #1255, opened 2018-01-26
Add 'tabindex=-1' to the featured image markup, to improve accessibil…

laurelfulford:

This is a reworking of #1243 based on feedback there. Rather than removing the `aria-hidden="true"` from Featured Images, it adds `tabindex="-1"

This update is to address missed feedback in #1158.

### DIFF
--- a/inc/template-tags.php
+++ b/inc/template-tags.php
@@ -132,7 +132,7 @@ function _s_post_thumbnail() {
 
 	<?php else : ?>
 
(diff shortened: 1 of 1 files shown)


grappler:
@laurelfulford It seems like this PR needs a refresh. Can you do it or should I?

laurelfulford:
Thanks for the heads up, @grappler! The PR has been refreshed.

PR #1262, opened 2018-02-19
Replace http with https in header.php

mauryaratan:
Replaced http with https, leading to mixed content warnings in server side scans.

### DIFF
--- a/header.php
+++ b/header.php
@@ -15,7 +15,7 @@
 <head>
 	<meta charset="<?php bloginfo( 'charset' ); ?>">
 	<meta name="viewport" content="width=device-width, initial-scale=1">
-	<link rel="profile" href="http://gmpg.org/xfn/11">
+	<link rel="profile" href="https://gmpg.org/xfn/11">
 
 	<?php wp_head(); ?>
 </head>


header.php
```diff
@@ -15,7 +15,7 @@
 <head>
 	<meta charset="<?php bloginfo( 'charset' ); ?>">
 	<meta name="viewport" content="width=device-width, initial-scale=1">
-	<link rel="profile" href="http://gmpg.org/xfn/11">
+	<link rel="profile" href="https://gmpg.org/xfn/11">
```
dingo-d: Can this be set to `//gmpg.org/xfn/11` (protocol relative URL)? This way you cover both `http` and `https` (locally you usually don't have any kind of certificate set up).

grappler: Sorry @dingo-d I don't agree that protocol relative URL is the best solution. The HTTPS link will always work regardless if your URL is HTTPS or not. https://www.paulirish.com/2010/the-protocol-relative-url/

edwardr:
Why not just run is_ssl() here to determine protocol?

grappler:
> Why not just run is_ssl() here to determine protocol?

There is no need for the extra complication. There is so little optimation from only running HTTP it is not worth the effort.

header.php
philiparthurmoore: @grappler is right.

mauryaratan: @grappler @philiparthurmoore Great! I've removed the commit, that leaves us with `https`. Good to merge :)

PR #1266, opened 2018-02-27
Two travis tweaks

jrfnl:
### Travis: speed up build times by disabling Xdebug

Xdebug is only **needed** when creating code coverage reports. Otherwise, it's just a nice extra, but a slow extra.
Disabling it will speed up the build times.

As suggested by John Blackbourn in https://johnblackbourn.com/reducing-travis-ci-build-times-for-wordpress-projects/

With an additional improvement to stabilize the command as suggested by Niklas Keller in  https://twitter.com/kelunik/status/954242454676475904

----
### Travis: stop build testing against HHVM, test against nightly

At some point in time, PHP was lagging behind and HHVM seemed like the future. What with PHP 7 being released and doing incredibly well, that time has passed and the popularity of HHVM is now marginal.

This commit drops testing against HHVM and starts testing against bleeding edge PHP nightly instead.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,11 +31,11 @@ matrix:
         - php: '7.0'
         # aliased to a recent 7.2.x version
         - php: '7.2'
-        # aliased to a recent hhvm version
-        - php: 'hhvm'
+        # bleeding edge PHP
+        - php: 'nightly'
 
     allow_failures:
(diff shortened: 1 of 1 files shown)


davidakennedy:
Makes sense! Thanks for the work here @jrfnl!

PR #1277, opened 2018-03-12
Wrap theme functions inside function_exists() check

pattonwebz:

This PR wraps theme functions inside of `function_exists()`.

NOTE: `/inc/woocommerse.php` functions are not currently wrapped.

#### Related issue(s): 
#1116

### DIFF
--- a/functions.php
+++ b/functions.php
@@ -83,53 +83,59 @@ if ( ! function_exists( '_s_setup' ) ) :
 endif;
 add_action( 'after_setup_theme', '_s_setup' );
 
-/**
- * Set the content width in pixels, based on the theme's design and stylesheet.
- *
- * Priority 0 to make it available to lower priority callbacks.
- *
- * @global int $content_width
- */
-function _s_content_width() {
-	// This variable is intended to be overruled from themes.
-	// Open WPCS issue: {@link https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/issues/1043}.
-	// phpcs:ignore WordPress.NamingConventions.PrefixAllGlobals.NonPrefixedVariableFound
-	$GLOBALS['content_width'] = apply_filters( '_s_content_width', 640 );
-}
+if ( ! function_exists( '_s_content_width' ) ) :
+	/**
+	 * Set the content width in pixels, based on the theme's design and stylesheet.
+	 *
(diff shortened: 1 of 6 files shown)


pattonwebz:
@davidakennedy The woocommerse.php file in the theme uses a different format for the `if` statements wrapping the function check than what was already used throughout the rest of the theme. That file uses `if () {..}` style wrappers where as the rest of the theme uses `if () : endif;`.

Should I normalize all the calls to be the same syntax and if so which method would be preferable?

samikeijonen:
I think current standard is to use `if () : endif;` in template files, and `if () {..}` in other files.

grappler:
As I mentioned in the issue https://github.com/Automattic/_s/issues/1116#issuecomment-307690097 I don't agree with making all functions pluggable.

pattonwebz:
@grappler so your preference is that only a small number (those which cannot be unhooked or support removed easily) should be overloadable?

I can do that if that's what is called for but you'll need to let me know what the list of functions to wrap are because people have different opinions on what is considered _easily removed_ lol

grappler:
> so your preference is that only a small number (those which cannot be unhooked or support removed easily) should be overloadable?

Yes that would be my preference.

> I can do that if that's what is called for but you'll need to let me know what the list of functions to wrap are because people have different opinions on what is considered easily removed lol

The steps that are needed would be https://github.com/Automattic/_s/issues/1116#issuecomment-307952674

I am not sure how to come to a decision or who makes the final decision..

philiparthurmoore:
Agree completely with @grappler's take on this.

Ismail-elkorchi:
Hi @pattonwebz 👋 ! Would you be still interested in updating this PR ? If not, no worries, I can pick up the work where you left off if you want, just let me know.

kangabell:
I'd find this change helpful!
@Ismail-elkorchi I see that @pattonwebz has not responded – are you still interested in picking this up?

Ismail-elkorchi:
@kangabell of course. I'll look it over today or tomorrow at the latest. thanks for the reminder :)

Ismail-elkorchi:
I have created a new PR https://github.com/Automattic/_s/pull/1480 to adress this issue based on https://github.com/Automattic/_s/issues/1116#issuecomment-307952674.

PR #1279, opened 2018-03-17
Fix datetime values

mboynes:

Use the date format `DATE_W3C` instead of `'c'` when populating `datetime` attributes. See trac [#20973](https://core.trac.wordpress.org/ticket/20973).

`get_the_date( 'c' )` will return the post date, which appears to be in local time (per the timezone setting), except that it appends `+00:00`. Therefore, it's technically the wrong time. Using `DATE_W3C` as the date format renders in the same format but appends the correct offset.


philiparthurmoore:
@grappler Looks good to me.

Issue #1278, opened 2018-03-13
hardcoded values in woocommerce boilerplate


smartcatdev:
In https://github.com/Automattic/_s/blob/master/inc/woocommerce.php#L78
```
/**
 * Products per page.
 *
 * @return integer number of products.
 */
function designr_woocommerce_products_per_page() {
    return 12;
}

add_filter( 'loop_shop_per_page', 'designr_woocommerce_products_per_page' );

/**
 * Product gallery thumnbail columns.
 *
 * @return integer number of columns.
 */
function designr_woocommerce_thumbnail_columns() {
    return 4;
}

add_filter( 'woocommerce_product_thumbnails_columns', 'designr_woocommerce_thumbnail_columns' );

/**
 * Default loop columns on product archives.
 *
 * @return integer products per row.
 */
function designr_woocommerce_loop_columns() {
    return 3;
}

add_filter( 'loop_shop_columns', 'designr_woocommerce_loop_columns' );
```

slushman:
Can you explain this issue a bit? I believe each filter is supposed to return a number, so I'm not sure what the problem is here.

nielslange:
@smartcatdev and @slushman: The `products per page` and the `products per row` can now be adjusted via the customizer.  

Furthermore, I wasn't able to figure out where to see `designr_woocommerce_thumbnail_columns()` in action. I've changed the return values from 1 to 10 but wasn't able to detect any change on the site. It looks like this function has no effect on this theme and could be removed.

PR #1281, opened 2018-03-19
Declare enhanced WooCommerce theme support

nielslange:

Added WooCommerce theme support as described in [Declaring WooCommerce support in themes](https://github.com/woocommerce/woocommerce/wiki/Declaring-WooCommerce-support-in-themes).

#### Related issue(s):

#1278

### DIFF
--- a/inc/woocommerce.php
+++ b/inc/woocommerce.php
@@ -16,7 +16,18 @@
  * @return void
  */
 function _s_woocommerce_setup() {
-	add_theme_support( 'woocommerce' );
+	add_theme_support( 'woocommerce', array(
+		'thumbnail_image_width' => 150,
+		'single_image_width'    => 300,
+		'product_grid'          => array(
+			'default_rows'    => 3,
+			'min_rows'        => 2,
+			'max_rows'        => 8,
+			'default_columns' => 4,
+			'min_columns'     => 2,
+			'max_columns'     => 5,
+		),
+	) );
 	add_theme_support( 'wc-product-gallery-zoom' );
 	add_theme_support( 'wc-product-gallery-lightbox' );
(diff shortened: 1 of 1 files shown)


tiagonoronha:
@nielslange thanks for the pull request! :)

One of the functions you removed, `_s_woocommerce_products_per_page()`, is used on a different function: https://github.com/nielslange/_s/blob/973876d14d62c8bfd16194a6ff0b7cc3228e1798/inc/woocommerce.php#L119

If you visit the Shop page with the changes in your pull request you'll get a Fatal Error:

`Fatal error: Uncaught Error: Call to undefined function _s_woocommerce_loop_columns() in /app/public/wp-content/themes/_s/inc/woocommerce.php`

You've a few options:

1) Add that function back, and do something like this: https://github.com/woocommerce/storefront/blob/master/inc/woocommerce/storefront-woocommerce-template-functions.php#L180-L196

2) Replace that function call with `wc_get_default_products_per_row()` instead.

3) Remove the wrappers completely from `_s`, as I'm doing in Storefront on this pull request: https://github.com/woocommerce/storefront/pull/779/files

As of WooCommerce 3.3, the column wrappers as not needed anymore because WooCommerce will add the number of columns to the wrapping div.

I suggest you go with option 3 and remove the wrapper completely, as that is the most future proof solution, but you might need to adjust the CSS if products are not displayed on a grid afterwards.

grappler:
@nielslange Is this something you still want to work on?

nielslange:
> @nielslange Is this something you still want to work on?

Sure, @grappler! I'm currently travelling and have limited time, but I'm happy to pick up this issue and finalise it.

nielslange:
Closed in favour of #1330

PR #1286, opened 2018-05-01
Try: Add AMP theme support

westonruter:
This is an exploration into adding `amp` theme support to Underscores. This new theme support flag is being introduced in v0.7 of the [AMP plugin](https://github.com/Automattic/amp-wp).

* Splits up script and style enqueueing since the former can be isolated to be prevented in AMP.
* Adds `_s_is_amp()` conditional function to check whether or not the response is being served as AMP .
* Prevents enqueueing `navigation.js`. Uses the `:focus-within` peudo class instead of the `focus` class which was added via JS. This pseudo-class is [supported](https://drafts.csswg.org/selectors-4/#the-focus-within-pseudo) by all browsers other than IE/Edge, and it ensures the submenu items are accessible even when JS is disabled. Also, [amp-bind](https://www.ampproject.org/docs/reference/components/amp-bind) is used to manage the toggled state of the navigation menu on mobile. To get amp-bind to work, the toggle `button` has to be switched over to a button-like label which is connected to a hidden checkbox.
* Prevent enqueueing `skip-link-focus-fix.js` since it is not allowed in AMP and it is only needed by IE11 anyway.
* Adds support for using `amp-live-list` for the comment list, allowing new comments to be added to the page without reloading. A refresh time of 60 seconds is used by default. A `_s_using_amp_live_list_comments()` conditional is added to check whether `amp-live-list` should be used in the template.
* The `meta[viewport]` is updated to add the AMP-required `minimum-scale=1` value.

With these changes applied to a theme, a vanilla WordPress install with the AMP plugin active can generate responses that do not trigger any AMP validation errors.

For more details on changes coming in AMP v0.7, see the [RC1 release post](https://make.xwp.co/2018/04/23/wordpress-amp-plugin-0-7-rc1/). See also another [full example theme](https://github.com/xwp/ampconf) which implements `amp` theme support, and a [WordPress install](https://ampdemo.xwp.io/) that has the theme active.


functions.php
```diff
@@ -24,6 +24,10 @@ function _s_setup() {
 		 */
 		load_theme_textdomain( '_s', get_template_directory() . '/languages' );
 
+		add_theme_support( 'amp', array(
+			'comments_live_list' => true,
+		) );
```
westonruter: Naturally this shouldn't be merged as-is into `master`. It would need to be opt-in by the theme developer.

functions.php
```diff
@@ -117,10 +121,56 @@ function _s_widgets_init() {
 add_action( 'widgets_init', '_s_widgets_init' );
 
 /**
- * Enqueue scripts and styles.
+ * Determine whether this is an AMP response.
+ *
+ * Note that this must only be called after the parse_query action.
+ *
+ * @link https://github.com/Automattic/amp-wp
```
westonruter: Missing `@return` here.

functions.php
```diff
@@ -117,10 +121,56 @@ function _s_widgets_init() {
...
+ */
+function _s_styles() {
 	wp_enqueue_style( '_s-style', get_stylesheet_uri() );
+}
+add_action( 'wp_enqueue_scripts', '_s_styles' );
+
+/**
+ * Enqueue scripts.
+ *
+ * This short-circuits in AMP because custom scripts are not allowed. There is are AMP equivalents provided elsewhere.
```
westonruter: `s/is are/are/`

header.php
```diff
@@ -44,8 +44,39 @@
...
+					class="menu-toggle"
+					for="site-navigation-expanded"
+					tabindex="0"
+					role="button"
+					aria-controls="primary-menu"
+					aria-expanded="false"
+					[aria-expanded]="siteNavigationMenu.expanded ? 'true' : 'false'"
+				>
+				<?php esc_html_e( 'Primary Menu', '_s' ); ?>
+			</label>
```
westonruter: Should be indented.

header.php
```diff
@@ -44,8 +44,39 @@
...
+							"expanded": false
+						}
+					</script>
+				</amp-state>
+				<input
+					id="site-navigation-expanded"
+					type="checkbox"
+					hidden
+					on="change:AMP.setState( { siteNavigationMenu: { expanded: event.checked } } )"
+				>
```
westonruter: This hidden checkbox is far from ideal. It's mainly here because AMP doesn't yet support a `toggleClass` action. See https://github.com/ampproject/amphtml/issues/1032

westonruter:
The mess I had originally proposed for using a `checkbox` instead of a `button` in AMP turned out to not be necessary, fortunately. It's much cleaner now.

comments.php
```diff
@@ -58,8 +67,23 @@
 		</ol><!-- .comment-list -->
 
 		<?php
+		if ( _s_using_amp_live_list_comments() ) {
+			add_filter( 'navigation_markup_template', '_s_add_amp_live_list_pagination_attribute' );
```
westonruter: This is a little awkward. It would be preferable if AMP allowed for a `nav` element (or `role="navigation"`) instead of requiring an element with a custom `pagination` attribute.

header.php
```diff
@@ -14,7 +14,7 @@
 <html <?php language_attributes(); ?>>
 <head>
 	<meta charset="<?php bloginfo( 'charset' ); ?>">
-	<meta name="viewport" content="width=device-width, initial-scale=1">
```
westonruter: Removal of space between values isn't actually required. This is just a bug in the AMP plugin fixed by https://github.com/Automattic/amp-wp/pull/1112

samikeijonen:
I didn't see an issue for AMP in `_s` so just checking in here. Looks like a nice job!

Couple of question:

- Now that Google, Automattic and XWP are pushing AMP plugin forward, I assume that decision for supporting AMP in `_s` have already been made?
- What does it actually mean. Is there going to be `is_amp()` checks and other AMP related stuff in template files or does this live in a different branch?

westonruter:
Thanks!

> Now that Google, Automattic and XWP are pushing AMP plugin forward, I assume that decision for supporting AMP in _s have already been made?

I don't think this decision has been made. However, in the same way that _s supports Jetpack and WooCommerce, it would seem logical to also support AMP.

> What does it actually mean. Is there going to be is_amp() checks and other AMP related stuff in template files or does this live in a different branch?

See https://github.com/Automattic/_s/pull/1286#discussion_r185347698. You're right that it should be opt-in. Maybe that looks like `add_theme_support('amp')` being commented-out by default, allowing a plugin author to enable it just by uncommenting the line?

grappler:
Thank you @westonruter for the pull request. It is nice to see what needs to be done on a themes side to make it compatable with the AMP Plugin.

> However, in the same way that _s supports Jetpack and WooCommerce, it would seem logical to also support AMP.

As WordPress.com supports AMP and _s is used as a base for WordPress.com themes then it may be logical. I don't think it is logical for themes built by the community for clients and WordPress.org.

Jetpack support can be easily be removed be deleting the `inc/jetpack.php`  file.
WooCommerce support is not included by default when using underscores.me.

If we were to add AMP support I would like to see easily removable.

- Move `_s_is_amp()` and `_s_using_amp_live_list_comments()` from `functions.php` to it's own file e.g. `inc/amp.php`.
- Create a separate comment template and load it via `comments_template()`.
- See if we can work with hooks in the `header.php`

header.php
```diff
@@ -44,8 +44,34 @@
...
+			<button
+				class="menu-toggle"
+				aria-controls="primary-menu"
+				aria-expanded="false"
+				<?php if ( _s_is_amp() ) : ?>
+					on="tap:AMP.setState( { siteNavigationMenu: { expanded: ! siteNavigationMenu.expanded } } )"
+					[aria-expanded]="siteNavigationMenu.expanded ? 'true' : 'false'"
+				<?php endif; ?>
+			>
+				<?php esc_html_e( 'Primary Menu', '_s' ); ?>
```
grappler: I could not figure out what the button is for. Is it for the mobile menu?

westonruter: Yes. If you resize the window to the mobile breakpoint then this button appears.

header.php
westonruter: I've restored the spaces since Automattic/amp-wp#1112 is closed. Can be tested with AMP plugin v1.0-alpha1.

westonruter:
@grappler I've simplified the PR by removing the `comments_live_list` support for now. I didn't think it would be good to have a separate template because then a theme author would have to maintain two separate comments templates.

Also, I've removed `amp` theme support from being hard-coded in `functions.php` entirely. This is because the AMP plugin as of [1.0-alpha1](https://make.xwp.co/2018/06/09/amp-plugin-release-v1-0-alpha1/) now includes an admin screen allowing the user to decide to what kind of AMP integration they want:

image

westonruter:
Note that [WP Rig](https://wprig.io/) has similar built-in AMP support now as well.

PR #1267, opened 2018-02-27
Bring code style up to the latest standards

jrfnl:
#### Changes proposed in this Pull Request:

This PR addresses code-style issues identified when testing the code against the WordPress Coding Standards.

Notes:
* Updates the versions of the various PHPCS libraries being used.
* Takes advantage of new features available in PHPCS 3.x and PHPCS 3.2+, such as parallel checking and improved inline whitelist annotations.
* Takes advantage of new sniffs and features available in the current stable version of WPCS (`0.14.1`).
* Builds upon work previously done by @grappler in #1157 for the actual code fixes.
* There is only one very tiny functional change (in commit 220615bc33aabe371b34a9f3d3df60692fe9c11e) where I've replaced an output escaping function with a more appropriate one.
    Other than that, the changes mainly concern whitespace and this PR will be easiest to review by using the `?w=1` trick when on the `files` tab of this PR.

I've made a new commit for each change to allow for properly documenting these changes in the commit history.

Please see the individual commit messages for (far) more detail.

#### Related issue(s):

Fixes #1253 
Closes #1157 (superseded) 
Closes #1186 (superseded)

PR #1291, opened 2018-05-17
Use composer to install PHPCS and friends on Travis

grappler:
Use composer instead of git to install WPCS and PHPCompatibility. This was recommend in a previous PR https://github.com/Automattic/_s/pull/1183#issuecomment-315844661

Changing the flow to https://github.com/Automattic/_s/pull/1267/commits/a0d2892cadbc290ac7e4eb507c7e5cfc55202a53

Only use the development version of PHPCompatibility. With this change we would use the stable version of PHPCS.  Due to an update to the PHPCS master the checks are failing. I would like to the necessary changes to happen along side the PHPCS and WPCS updates instead of having to do them with other changes.

#### Related issue(s):
#1267

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -48,19 +48,13 @@ matrix:
 before_script:
     # Speed up build time by disabling Xdebug.
     - phpenv config-rm xdebug.ini || echo 'No xdebug config.'
-    # Set up temporary paths.
-    - export PHPCS_DIR=/tmp/phpcs
-    - export WPCS_DIR=/tmp/wpcs
-    - export PHPCOMPAT_DIR=/tmp/phpcompatibility
-    # Install CodeSniffer for WordPress Coding Standards checks.
-    - if [[ "$SNIFF" == "1" ]]; then git clone -b master --depth 1 https://github.com/squizlabs/PHP_CodeSniffer.git $PHPCS_DIR; fi
-    # Install WordPress Coding Standards.
-    - if [[ "$SNIFF" == "1" ]]; then git clone -b 0.14.1 --depth 1 https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards.git $WPCS_DIR; fi
-    # Install PHP Compatibility sniffs.
-    - if [[ "$SNIFF" == "1" ]]; then git clone -b master --depth 1 https://github.com/wimg/PHPCompatibility.git $PHPCOMPAT_DIR; fi
+    # Install WordPress Coding Standards & PHP Compatibility sniffs.
+    - if [[ "$SNIFF" == "1" ]]; then composer global require wp-coding-standards/wpcs:~0.14.1 wimg/php-compatibility:dev-master --prefer-stable --no-interaction --no-suggest; fi
+    # Set composer PATH.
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -48,19 +48,13 @@ matrix:
...
-    - export WPCS_DIR=/tmp/wpcs
-    - export PHPCOMPAT_DIR=/tmp/phpcompatibility
-    # Install CodeSniffer for WordPress Coding Standards checks.
-    - if [[ "$SNIFF" == "1" ]]; then git clone -b master --depth 1 https://github.com/squizlabs/PHP_CodeSniffer.git $PHPCS_DIR; fi
-    # Install WordPress Coding Standards.
-    - if [[ "$SNIFF" == "1" ]]; then git clone -b 0.14.1 --depth 1 https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards.git $WPCS_DIR; fi
-    # Install PHP Compatibility sniffs.
-    - if [[ "$SNIFF" == "1" ]]; then git clone -b master --depth 1 https://github.com/wimg/PHPCompatibility.git $PHPCOMPAT_DIR; fi
+    # Install WordPress Coding Standards & PHP Compatibility sniffs.
+    - if [[ "$SNIFF" == "1" ]]; then composer global require wp-coding-standards/wpcs:~0.14.1 wimg/php-compatibility:dev-master --prefer-stable --no-interaction --no-suggest; fi
```
jrfnl: Please add `dealerdirect/phpcodesniffer-composer-installer:^0.4.3`. This is a Composer plugin which will sort out the installed paths automatically and is suggested for that reason by both the standards.

.travis.yml
```diff
@@ -48,19 +48,13 @@ matrix:
...
+    # Set composer PATH.
+    - if [[ "$SNIFF" == "1" ]]; then export PATH=$HOME/.composer/vendor/bin:$PATH; fi
     # Set install path for PHPCS sniffs.
     # @link https://github.com/squizlabs/PHP_CodeSniffer/blob/4237c2fc98cc838730b76ee9cee316f99286a2a7/CodeSniffer.php#L1941
-    - if [[ "$SNIFF" == "1" ]]; then $PHPCS_DIR/bin/phpcs --config-set installed_paths $WPCS_DIR,$PHPCOMPAT_DIR; fi
+    - if [[ "$SNIFF" == "1" ]]; then phpcs --config-set installed_paths $HOME/.composer/vendor/wp-coding-standards/wpcs,$HOME/.composer/vendor/wimg/php-compatibility; fi
```
jrfnl: If the above mentioned Composer plugin is added, this line can be removed completely.

jrfnl:
Interesting approach doing this without a `composer.json`, though I do think it would be better to have a `composer.json` file, if only to document which versions of the various libraries are being used in a more obvious way than hidden away in a Travis script.

Regarding the build failure: this has to do with a change in PHPCS - see https://github.com/squizlabs/PHP_CodeSniffer/issues/1539. Please note that to version which would currently be used `3.2.3` is buggy regarding the new PHPCS annotations.

Ismail-elkorchi:
Closing as composer was added in https://github.com/Automattic/_s/pull/1386.

Issue #1292, opened 2018-05-18
Add PostCSS Partials


cedon:
I have created a refactored set of SASS partials to allow developers to utilize _s in a PostCSS workflow. Since PostCSS allows access to tools not necessarily available to SASS out of the box, like CSS Next, I thought this might be an option to include in repository.

Pull request is being submitted from my fork.

mtomas7:
I based my theme on _s and currently I'm evaluating the option to use PostCSS and this is a great help to start! Thank you!

Ismail-elkorchi:
Hello @cedon ! Thank you for the issue report and the Pull Request.

While PostCSS offer more flexibility thanks to all the available plugins, it depend on more moving parts and its maintenance is more complex. It's much simpler with SASS since it depend only on `node-sass`.

PR #1293, opened 2018-05-18
Post css

cedon:
Created PostCSS partials based on the SASS files. This would allow theme developers to use _s out of the box with minimal changes to workflow.

#### Related issue(s):
#1292


Ismail-elkorchi:
see https://github.com/Automattic/_s/issues/1292#issuecomment-612685108

PR #1297, opened 2018-06-03
Add basic composer.json file.

alexmacarthur:
I'd be helpful to be able to require the _s theme via Composer using the [Composer Installers](https://github.com/composer/installers) package, which would automatically place the `wordpress-theme` type in the specified themes directory. 

An example of how someone might leverage this Composer file can be seen here: 
https://github.com/alexmacarthur/wp-skateboard/tree/starter-theme-underscores

As a solution to all of this, I've created a very basic `composer.json` file. It can certainly be improved, but contains all the essential information as a starting point.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,0 +1,7 @@
+{
+    "name": "_s",
+    "description": "Hi. I'm a starter theme called '_s', or 'underscores', if you like. I'm a theme meant for hacking so don't use me as a parent theme. Instead try turning me into the next, most awesome, WordPress theme out there. That's what I'm here for.",
+    "type": "wordpress-theme",
+    "license": "GPL-2.0 or later",
+    "version": "1.0.0"
+}


grappler:
Thank you @alexmacarthur for the PR.

We were discussing in #1291 about adding a `composer.json` file to define the dependencies on tools like PHPCS.

There are three main groups of uses for `_s`
1. Themes for WordPress.com
1. Themes for WordPress.org
1. Custom Themes for clients

The `composer.json` is only useful for the last group as WordPress.com does not use `composer`(as far as I know). For WordPress.org there is https://wpackagist.org/

One question on my mind is:
> What happens to the `composer.json` when the theme is generated from [underscores.me](https://underscores.me/)?

I see two options.

1. Not include it in the final zip
1. Include it in the final zip but make sure the name is changed correctly
1. Not to include it by default but allow an option for it.

I don't think we should include the `composer.json` by default as most people would not delete it. We could look into adding a new option to underscores.me but that will take time. Previous discussion can be found in https://github.com/Automattic/underscores.me/issues/31

alexmacarthur:
Thanks for the feedback, @grappler. Agreed, the value for something like this would be found by people using Composer in their development, and I don't think it'd be worth including in the underscores.me generator. 

Would there be any harm in including the file in the repo anyway, and then having people do what they want with it after that? And if there is harm, would the value outweigh that harm? 

I'm obviously partial to having a Composer file included in the repo itself, but I'm also not aware of the scope of what it might impact.

tomjn:
I would consider adding the file, and changing the package type to `project`, people aren't meant to install underscores as a theme, but to use it as a starter project

Otherwise I would add the composer installers package as a dependency

Ismail-elkorchi:
@alexmacarthur Thank you for the Pull Request. Closing as `composer.json` file was added in https://github.com/Automattic/_s/pull/1386.

Issue #1289, opened 2018-05-13
Remove aria-expanded on menu UL element?


samikeijonen:
Just leaving note here before I forget. We should check do we really need `aria-expanded` on menu (`<ul>`) itself. I'm thinking we don't need it.

[Here is the JS line for it](https://github.com/Automattic/_s/blob/master/js/navigation.js#L28).

mrwweb:
Unless this was included due to some type of AT implementing the spec wrong, it's clear that [aria-expanded only goes on the control, not the container](https://www.w3.org/WAI/GL/wiki/Using_the_WAI-ARIA_aria-expanded_state_to_mark_expandable_and_collapsible_regions#Example_1:_Using_a_button_to_collapse_and_expand_a_region).

There are also [two](https://github.com/Automattic/_s/blob/master/js/navigation.js#L37) other [lines](https://github.com/Automattic/_s/blob/master/js/navigation.js#L41) that would need to be changed.

jaredcaraway:
I did some research on this today, and I agree with @samikeijonen and @mrwweb - unless I'm overlooking something, I don't see anything supporting the need for the `aria-expanded` attribute on the menu `<ul>` itself.

[From WAI-ARIA Authoring Practices 1.1](https://www.w3.org/TR/wai-aria-practices/#menubutton):

> - The element that opens the menu has role button.
> - The element with role button has aria-haspopup set to either menu or true.
> - When the menu is displayed, the element with role button has aria-expanded set to true. When the menu is hidden, it is recommended that aria-expanded is not present. If aria-expanded is specified when the menu is hidden, it is set to false.
> - The element that contains the menu items displayed by activating the button has role menu.
> - Optionally, the element with role button has a value specified for aria-controls that refers to the element with role menu.
> - Additional roles, states, and properties needed for the menu element are described in 3.14 Menu or Menu bar.

There is also a [Navigation Menu Button Example](https://www.w3.org/TR/wai-aria-practices/examples/menu-button/menu-button-links.html) which shows this in action - the `aria-expanded` attribute is only applied to the button element.
![screenshot from 2018-06-27 16-18-45](https://user-images.githubusercontent.com/436100/42000286-d75ee1e0-7a25-11e8-8f21-d9fc816be4d4.png)

It's worth noting some differences from the WAI-ARIA doc:
- It is recommended to _remove_ the `aria-expanded` attribute when the element is not expanded rather than setting its value to false, but that approach is not explicitly incorrect
[...]

PR #1303, opened 2018-06-27
Remove unnecessary aria-expanded attribute on ul

jaredcaraway:
Remove `aria-expanded` attribute from `<ul>` element, as it should only be applied to the button controlling the showing/hiding of the menu. (See [my comment](https://github.com/Automattic/_s/issues/1289#issuecomment-400845419) on #1289 for more detail.)

#### Related issue(s):
Fixes https://github.com/Automattic/_s/issues/1289

### DIFF
--- a/js/navigation.js
+++ b/js/navigation.js
@@ -25,7 +25,6 @@
 		return;
 	}
 
-	menu.setAttribute( 'aria-expanded', 'false' );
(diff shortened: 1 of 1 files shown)


Ismail-elkorchi:
I removed one more instance of `aria-expanded` from the`<ul>` element that was introduced after the merging of https://github.com/Automattic/_s/pull/1379, and rebased the branch on master so the Travis test suite pass.

PR #1311, opened 2018-08-09
CS/QA update for WPCS 1.0.0, PHPCS 3.3.1 & PHPCompatibility

jrfnl:
## Changes proposed in this Pull Request:

This PR updates the CS/QA related toolset to use the latest versions of these tools and updates the code-base to comply.

### QA: Break stylesheets out of browser cache on each new release of a theme

_WPCS 1.0.0 contains a new sniff which identifies this issue, which is why it has to be solved for the update to be possible_

**Type: Functional change**

The `wp_enqueue_style()` function allows for passing a `$version` parameter which is used to force browsers to load a fresh copy of a stylesheet when it has been changed.

If this parameter is not passed, the version number will be set to the version of the WP install which doesn't make sense at all for a theme which is not shipped with WP itself.

This commit adds a `_S_VERSION` constant which should be changed to the current version number of the actual finished theme on each new release.

This constant is subsequently used in all `wp_enqueue_style()` function calls to ensure that on each new release of the theme, the browser cache is cleared for these stylesheets.

Ref: https://developer.wordpress.org/reference/functions/wp_enqueue_style/

---

### CS: Minor fixes to comply with WPCS 1.0.0 / PHPCS 3.3.1

**Type: CS update**

Nothing very interesting, just some minor fixes to comply with the latest versions of all sniffs being tested against.

---

### Use PHPCompatibilityWP

**Type: Build tools**

Since mid July, the PHPCompatibility project offers dedicated CMS rulesets which already account for all back-fills/poly-fills.

This commit removes the `_s` native version of that ruleset in favour of using this dedicated WP-based ruleset.

Additionally, the PHPCompatibility repo has moved to a dedicated GitHub organisation, so that change is included here as well.

Refs:
* https://github.com/PHPCompatibility/PHPCompatibility/releases/tag/8.2.0
* https://github.com/phpcompatibility/phpcompatibilitywp

**Note**: I would strongly recommend installing these dependencies via Composer instead. @grappler has created a good setup for this in PR #1291 which I would recommend for merging, though it will need to be updated to be in line with the changes made in this PR.

---

### Use WPCS 1.0.0 and sniff against a high PHP version for the fastest and most reliable results.

**Type: Build tools**

No ruleset changes are needed.

Refs: https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/releases/tag/1.0.0

### DIFF
--- a/functions.php
+++ b/functions.php
@@ -7,6 +7,11 @@
  * @package _s
  */
 
+if ( ! defined( '_S_VERSION' ) ) {
+	// Replace #.# with the version number of the theme on each release.
+	define( '_S_VERSION', '#.#' );
+}
+
 if ( ! function_exists( '_s_setup' ) ) :
 	/**
 	 * Sets up theme defaults and registers support for various WordPress features.
@@ -120,11 +125,11 @@ add_action( 'widgets_init', '_s_widgets_init' );
  * Enqueue scripts and styles.
  */
 function _s_scripts() {
-	wp_enqueue_style( '_s-style', get_stylesheet_uri() );
+	wp_enqueue_style( '_s-style', get_stylesheet_uri(), array(), _S_VERSION );
 
-	wp_enqueue_script( '_s-navigation', get_template_directory_uri() . '/js/navigation.js', array(), '20151215', true );
+	wp_enqueue_script( '_s-navigation', get_template_directory_uri() . '/js/navigation.js', array(), _S_VERSION, true );
 
-	wp_enqueue_script( '_s-skip-link-focus-fix', get_template_directory_uri() . '/js/skip-link-focus-fix.js', array(), '20151215', true );
+	wp_enqueue_script( '_s-skip-link-focus-fix', get_template_directory_uri() . '/js/skip-link-focus-fix.js', array(), _S_VERSION, true );
 
 	if ( is_singular() && comments_open() && get_option( 'thread_comments' ) ) {
 		wp_enqueue_script( 'comment-reply' );
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,11 +26,11 @@ matrix:
           dist: precise
         # aliased to a recent 5.6.x version
         - php: '5.6'
-          env: SNIFF=1
         # aliased to a recent 7.0.x version
         - php: '7.0'
         # aliased to a recent 7.2.x version
         - php: '7.2'
+          env: SNIFF=1
         # bleeding edge PHP
         - php: 'nightly'
 
@@ -56,7 +56,7 @@ before_script:
     # Install CodeSniffer for WordPress Coding Standards checks.
     - if [[ "$SNIFF" == "1" ]]; then git clone -b master --depth 1 https://github.com/squizlabs/PHP_CodeSniffer.git $PHPCS_DIR; fi
     # Install WordPress Coding Standards.
-    - if [[ "$SNIFF" == "1" ]]; then git clone -b 0.14.1 --depth 1 https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards.git $WPCS_DIR; fi
+    - if [[ "$SNIFF" == "1" ]]; then git clone -b 1.0.0 --depth 1 https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards.git $WPCS_DIR; fi
     # Install PHP Compatibility sniffs.
     - if [[ "$SNIFF" == "1" ]]; then git clone -b master --depth 1 https://github.com/PHPCompatibility/PHPCompatibility.git $PHPCOMPAT_DIR; fi
     # Install WP PHP Compatibility ruleset.
--- a/.travis.yml
+++ b/.travis.yml
@@ -52,15 +52,18 @@ before_script:
     - export PHPCS_DIR=/tmp/phpcs
     - export WPCS_DIR=/tmp/wpcs
     - export PHPCOMPAT_DIR=/tmp/phpcompatibility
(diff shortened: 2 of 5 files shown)


jrfnl:
Hmm... no response at all and in the mean time WPCS 1.1.0 has come out... makes it difficult to keep up.

samikeijonen:
If I was in charge I'd say go for it! Looks good.

jrfnl:
I'm willing to update this PR to account for WPCS 1.1.0 as well as PHPCompatibilityWP 2.0.0, however, I can't be really be bothered if this PR is left open to rot....

philiparthurmoore:
This looks good.

jrfnl:
> This looks good.

Except the build won't pass anymore now WPCS 1.1.0 has come out...

philiparthurmoore:
Then we should update it. 😝

limestreet:
Hi @jrfnl 
I see **PHPCompatibilityWP** uses **PHPCompatibilityParagonieRandomCompat**:
https://github.com/PHPCompatibility/PHPCompatibilityWP/blob/master/PHPCompatibilityWP/ruleset.xml#L34
It looks like it should be added here as well (as you added it in **PHPCompatibilityWP 2.0**):
https://github.com/Automattic/_s/pull/1311/files#diff-354f30a63fb0907d4ad57269548329e3

jrfnl:
@limestreet I'm aware - see: https://github.com/Automattic/_s/pull/1311#issuecomment-427703801

functions.php
```diff
@@ -120,11 +125,11 @@ function _s_widgets_init() {
  * Enqueue scripts and styles.
  */
 function _s_scripts() {
-	wp_enqueue_style( '_s-style', get_stylesheet_uri() );
+	wp_enqueue_style( '_s-style', get_stylesheet_uri(), array(), _S_VERSION );
 
-	wp_enqueue_script( '_s-navigation', get_template_directory_uri() . '/js/navigation.js', array(), '20151215', true );
+	wp_enqueue_script( '_s-navigation', get_template_directory_uri() . '/js/navigation.js', array(), _S_VERSION, true );
```
grappler: Can we leave the version as is? The script does not get changed with every theme version. This way we would not need the constant.

jrfnl: Except that the scripts have been changed a couple of times already and people keep forgetting to update this cache-busting number, so I would *strongly* recommend leaving it as is (with the constant), as refreshing the script once on every new release is a minor slow-down, while serving an out-of-date script can break a site.

grappler: I don't agree but I don't want to stand in the way of this PR being merged. We might need a PR for https://github.com/Automattic/underscores.me/ too to search and replace `_S`.

enkrates:
The build is failing on a missing dependency, the [Paragonie compat sniff](https://github.com/PHPCompatibility/PHPCompatibilityParagonie). It seems like it'd be pretty simple to also install that along the other sniffs, I'd be happy to help if that'd keep this PR moving forward.

jrfnl:
@enkrates The build still wouldn't pass, but that is because no-one of the maintainers bothers to merge in a timely manner.
See https://github.com/Automattic/_s/pull/1311#issuecomment-427703801 and https://github.com/Automattic/_s/pull/1311#issuecomment-438033508

qwerbycom:
Where are the maintainers of this build?

crunnells:
@jrfnl If you'd like to update this PR to pass the build and update the version of WPCS I'll get it merged in a timely manner.

philiparthurmoore:
@jrfnl I haven't kept up with this much. Are you done with it and just waiting for it to be merged? Trying to get a sense of current status without reading the entire thread. It seems like it was ready a while back and then nothing happened.

jrfnl:
@philiparthurmoore TL:DR: This PR needs a big update as it was left to rot for a year and a half and all the standards involved have moved on since then.

I'd be willing to do the update, but only if I have a solid commitment that the PR will be reviewed & merged within two weeks of doing the update as otherwise it will quickly be out of date again.

_Note: By now, I have very little faith in anything being merged within a reasonable timespan in this project based on (the lack of) progress on any PR in the past two years._

Also: it would be _so_ much better if this project would use Composer to manage the (dev) dependencies as that way, it won't be bound to `dev-master` versions of the CS dependencies which will break builds unexpectedly.

From the PR description above:
> Note: I would strongly recommend installing these dependencies via Composer instead. @grappler has created a good setup for this in PR #1291 which I would recommend for merging, though it will need to be updated to be in line with the changes made in this PR.

philiparthurmoore:
@jrfnl, To be honest, I don't see the point of submitting another PR unless there is a firm commitment from the Automattic theme team to revive _s. Otherwise, I consider it in maintenance mode until I see activity that proves otherwise. I have the rights to merge pretty much anything and everything regarding _s but it doesn't make any sense for me to be the one doing the work if Automattic is using _s (or not) for themes now. Dave mentioned that they were using another starter theme a while back so I'm still a little confused about who leads this now.

szepeviktor:
@jrfnl Begetting a child and being pregnant takes ~50% of time this PR gets merged.

jrfnl:
@szepeviktor I'd still rather send in a PR though ;-)

Ismail-elkorchi:
In https://github.com/Automattic/_s/pull/1386, I updated the CS tools and made Travis use these, then I created https://github.com/Automattic/_s/pull/1388 to make the theme comply with WPCS.

Ismail-elkorchi:
@jrfnl Thank you for the Pull Request. The issues regarding coding standards and Travis were resolved in https://github.com/Automattic/_s/pull/1388 and https://github.com/Automattic/_s/pull/1386.

PR #1320, opened 2018-09-22
Unscramble echo to readable printf

szepeviktor:
Is it OK?

Use printf.


szepeviktor:
You're welcome.

Issue #1206, opened 2017-09-19
Is skip-link-focus-fix.js still necessary?


startover909:
All the information I can find regarding this script and the bug it's intended to fix are years old. I tested current versions of Chrome FireFox IE11 and Edge and none exhibit the bug. Maybe it was fixed a long time ago? Or does it only affect some very old versions of browsers with very few users (and I'd imagine that even fewer keyboard-only users would choose to use such browsers).  So is it safe to let this go at this point of time in 2017?

samikeijonen:
It's still an issue in IE11.

startover909:
Hmmm, I can't reproduce it on IE 11 in Windows 10 v.1703.  Skip to content link then tab key worked as intended without the fix js.  Maybe I'm running a different theme that's somehow different.

samikeijonen:
On WIN8/IE11 I can still see this issue. My computer wants to update to WIN10 but I try not to:)

startover909:
I actually have access to a Win 8.1 system too and the IE version on there is 11.0.9600.18793 with Update Version 11.0.46 (KB4036586), which does not exhibit the issue with.  Test site:

https://www.liujack.me

Do you have a link to a site which shows the bug for you so that I can test it as well?

Thanks.

samikeijonen:
I don't see the issue on that site but I don't have time to debug the theme used in there.

However I just tested with Twenty Seventeen, vanilla _s, and couple of my own themes. They all have the issue if I remove the skip-link JS script.

Can you test with Twenty Seventeen and vanilla `_s` theme also?

davidakennedy:
Hi @startover909! Thanks for creating this issue. That JavaScript is still needed as @samikeijonen noted, the issue persists in IE 11. I tested on Browserstack with Windows 10 and IE 11 and Windows 8.1 and IE 11, and it's still a problem.

You should see the issue if you remove the script and test. The focus will not move to the content area when you click the skip to content link after "tabbing" to it. Instead it goes back tot he top of the page.

I'm going to close this issue, but feel free to comment again if I'm missing something.

Salamander3:
Well it's a dequeue for me.

westonruter:
[...]

PR #1323, opened 2018-10-10
Print skip-link-focus-fix inline instead of enqueueing as blocking script

westonruter:

Enqueueing scripts is very costly since the browser will block rendering until the script is loaded. What's more is that it is only applicable in IE11, so it is particularly bad to have this blocking script loaded for 90% of browsers which will never use it. (If IE11 supported conditional comments, those could have been used here, but it does not.) Blocking scripts should be minimized as much as possible, either by adding the `async` attribute or by outputting the `script` inline.

While core doesn't yet have built-in support for `async`, core does currently add an inline script for adding the emoji logic (`print_emoji_detection_script()`/`_print_emoji_detection_script()`). A similar approach is proposed here for the `skip-link-focus-fix` script. Just like for the emoji detection script, the skip-link-focus-fix is small in size and it is only applicable to IE11.

#### Related issue(s):

#1206

### DIFF
--- a/functions.php
+++ b/functions.php
@@ -124,13 +124,26 @@ function _s_scripts() {
 
 	wp_enqueue_script( '_s-navigation', get_template_directory_uri() . '/js/navigation.js', array(), '20151215', true );
 
-	wp_enqueue_script( '_s-skip-link-focus-fix', get_template_directory_uri() . '/js/skip-link-focus-fix.js', array(), '20151215', true );
-
 	if ( is_singular() && comments_open() && get_option( 'thread_comments' ) ) {
 		wp_enqueue_script( 'comment-reply' );
 	}
 }
 add_action( 'wp_enqueue_scripts', '_s_scripts' );
+
+/**
+ * Fix skip link focus in IE11.
+ *
+ * This does not enqueue the script because it is tiny and because it is only for IE11,
+ * thus it does not warrant having an entire dedicated blocking script being loaded.
+ *
+ * @link https://git.io/vWdr2
+ */
+function _s_skip_link_focus_fix() {
+	echo '<script>';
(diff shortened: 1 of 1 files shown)


functions.php
```diff
@@ -124,14 +124,27 @@ function _s_scripts() {
...
+ * Fix skip link focus in IE11.
+ *
+ * This does not enqueue the script because it is tiny and because it is only for IE11,
+ * thus it does not warrant having an entire dedicated blocking script being loaded.
+ *
+ * @link https://git.io/vWdr2
+ */
+function _s_skip_link_focus_fix() {
+	echo '<script>';
+	echo file_get_contents( get_template_directory() . '/js/skip-link-focus-fix.js' );
```
samikeijonen: Is `file_get_contents ` still banned from wp.org? If it is, can you push to get it allowed.

joshmcrty:
It should be possible to add an async (or defer) attribute using the `script_loader_tag` filter. See https://developer.wordpress.org/reference/hooks/script_loader_tag/ and https://matthewhorne.me/defer-async-wordpress-scripts/ for an example.

westonruter:
@joshmcrty it is indeed possible. However, `defer` at least is still not ideal. As noted in the corresponding PR https://github.com/wprig/wprig/pull/139:

> While WP Rig is using defer this still causes a performance penalty since it blocks `DOMContentLoaded` from firing. So printing the script inline is a better approach.

It's true that `async` avoids this issue. Nevertheless, avoiding an extra HTTP request for such a small script seems beneficial as in the case of the emoji detection script.

Ismail-elkorchi:
@westonruter Thank you for the pull request. Can you please update it to include the changes that were made in https://github.com/WordPress/twentynineteen/pull/47 ?

PR #1330, opened 2018-11-14
Declare enhanced WooCommerce theme support

### LINKED ISSUE #1278: hardcoded values in woocommerce boilerplate

nielslange:
Fixes #1278 

Remove hard-coded product grid values and handle product grid values by the Customizer instead.

#### Related issue(s):

n/a

### DIFF
--- a/inc/woocommerce.php
+++ b/inc/woocommerce.php
@@ -16,7 +16,18 @@
  * @return void
  */
 function _s_woocommerce_setup() {
-	add_theme_support( 'woocommerce' );
+	add_theme_support( 'woocommerce', array(
+		'thumbnail_image_width' => 150,
+		'single_image_width'    => 300,
+		'product_grid'          => array(
+			'default_rows'    => 3,
+			'min_rows'        => 2,
+			'max_rows'        => 8,
+			'default_columns' => 4,
+			'min_columns'     => 2,
+			'max_columns'     => 5,
+		),
+	) );
 	add_theme_support( 'wc-product-gallery-zoom' );
 	add_theme_support( 'wc-product-gallery-lightbox' );
 	add_theme_support( 'wc-product-gallery-slider' );
@@ -71,36 +82,6 @@ function _s_woocommerce_active_body_class( $classes ) {
 add_filter( 'body_class', '_s_woocommerce_active_body_class' );
 
 /**
- * Products per page.
- *
- * @return integer number of products.
- */
-function _s_woocommerce_products_per_page() {
-	return 12;
-}
-add_filter( 'loop_shop_per_page', '_s_woocommerce_products_per_page' );
-
-/**
- * Product gallery thumnbail columns.
- *
- * @return integer number of columns.
- */
-function _s_woocommerce_thumbnail_columns() {
-	return 4;
-}
-add_filter( 'woocommerce_product_thumbnails_columns', '_s_woocommerce_thumbnail_columns' );
-
-/**
- * Default loop columns on product archives.
- *
- * @return integer products per row.
- */
-function _s_woocommerce_loop_columns() {
-	return 3;
-}
-add_filter( 'loop_shop_columns', '_s_woocommerce_loop_columns' );
-
-/**
  * Related Products Args.
  *
  * @param array $args related products args.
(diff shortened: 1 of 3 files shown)


nielslange:
@grappler Would you mind taking a look at this PR to see if it's in a shippable state?

inc/woocommerce.php
```diff
@@ -16,7 +16,18 @@
  * @return void
  */
 function _s_woocommerce_setup() {
```
dingo-d: I couldn't comment on the `add_action` line, but just a small comment: would it be good to wrap the `add_action( 'after_setup_theme', '_s_woocommerce_setup' )` in a conditional like:

```php
if ( class_exists( 'WooCommerce' ) ) {
    add_action( 'after_setup_theme', '_s_woocommerce_setup' )
}
```

So that this function is added only in the case the `WooCommerce` plugin is installed and active?

grappler: That is not necessary as it is already being done on a file level. See: https://github.com/Automattic/_s/blob/master/functions.php#L165

inc/woocommerce.php
```diff
@@ -16,7 +16,18 @@
...
-	add_theme_support( 'woocommerce' );
+	add_theme_support( 'woocommerce', array(
+		'thumbnail_image_width' => 150,
+		'single_image_width'    => 300,
+		'product_grid'          => array(
+			'default_rows'    => 3,
+			'min_rows'        => 2,
```
grappler:
```suggestion
			'min_rows'        => 1,
```
There should be no reason not to allow a single row. This is the WooCommerce default.

inc/woocommerce.php
```diff
@@ -16,7 +16,18 @@
...
+			'max_rows'        => 8,
```
grappler:
```suggestion
			'max_rows'        => '',
```
Same here. We should not restrict the maximum rows. This is the WooCommerce default.

inc/woocommerce.php
```diff
@@ -16,7 +16,18 @@
...
+			'default_columns' => 4,
+			'min_columns'     => 2,
```
grappler:
```suggestion
			'min_columns'     => 1,
```
There should be no reason not to allow a single column. This is the WooCommerce default.

inc/woocommerce.php
```diff
@@ -16,7 +16,18 @@
...
+			'max_columns'     => 5,
```
grappler:
```suggestion
			'max_columns'     => 6,
```
As we support 6 columns in the CSS.

inc/woocommerce.php
dingo-d: Oh, didn't notice this. Great :+1:

Ismail-elkorchi:
@nielslange are you still interested in updating this PR? I  can pick up the work where you left off if you want.

nielslange:
> @nielslange are you still interested in updating this PR? I can pick up the work where you left off if you want.

Hi @Ismail-elkorchi 👋 Feel free to take over! 😉

Ismail-elkorchi:
@nielslange Thank you for the quick reply 😀.

nielslange:
@Ismail-elkorchi Thank you for pinking up this issue! 🙌

Ismail-elkorchi:
Closing in favour of https://github.com/Automattic/_s/pull/1394.

PR #1335, opened 2018-12-03
Add actions to header.php

tazotodua:
I think it's very good practice to suggest themes to have `do_action` hooks as many place as possible. So, end-users won't have to manually modify themes.

#### Related issue(s):


grappler:
Closing. See #1340

PR #1340, opened 2019-01-10
Missing hooks everywhere...

ttodua:
I was initially surprised how Automattic missed the most important & educational parts of any theme/plugin. Hooks - that's what everything should use. So, it's quite negatively affected, when theme/plugin developers are not educated or encourage to implement hooks everywhere. So, let's do that.

#### Related issue(s):

Missing hooks everywhere.

### DIFF
--- a/archive.php
+++ b/archive.php
@@ -9,7 +9,7 @@
 
 get_header();
 ?>
-	<?php do_action("_s_archive_top");?>
+	<?php do_action("_s_archive_head");?>
 
 	<div id="primary" class="content-area">
 		<main id="main" class="site-main">
@@ -53,7 +53,7 @@ get_header();
 		</main><!-- #main -->
 	</div><!-- #primary -->
 
-	<?php do_action("_s_archive_bottom");?>
+	<?php do_action("_s_archive_foot");?>
 
 <?php
 get_sidebar();
--- a/archive.php
+++ b/archive.php
@@ -9,12 +9,12 @@
 
 get_header();
 ?>
-	<?php do_action("_s_archive_head");?>
+	<?php do_action( "_s_archive_head" );?>
 
 	<div id="primary" class="content-area">
 		<main id="main" class="site-main">
 
-		<?php do_action("_s_archive_in_primary");?>
+		<?php do_action( "_s_archive_in_primary" );?>
 			
 		<?php if ( have_posts() ) : ?>
 
@@ -25,7 +25,7 @@ get_header();
 				?>
 			</header><!-- .page-header -->
 
-			<?php do_action("_s_archive_middle");?>
+			<?php do_action( "_s_archive_middle" );?>
 			
 			<?php
 			/* Start the Loop */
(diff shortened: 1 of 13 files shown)


sumnermic:
I am wondering what other people decide their theme ought to be? I personally like the use of hooks within WooCommerce as it helps tidy stuff up. 

But will it be necessary to do so with a theme?

tazotodua:
I don't remember a site I've build, and there wasn't a need to place a custom function somewhere in a there. So, we had always to "modify" the current theme code to inject one line codes of "do_action" in different parts. All themes should have it, why only woocommerce? what if we want to change other themes too?

dingo-d:
I'm more used to seeing hooks in plugins than in themes but I guess I don't mind having them in a theme 😄

grappler:
There was a previous pull request https://github.com/Automattic/_s/pull/671 to add the support for "[Theme Hook Alliance"](https://github.com/zamoose/themehookalliance). It was decided again merging those hooks.

As `_s` is meant to be adapted and not necessarily extended I don't think the hooks are needed.

I would rather see core support a standard set of the hooks https://core.trac.wordpress.org/ticket/21506

As most templates are small then can be easily be overridden with causing too much code to duplicated.

There is not need for the `_s_*_bloginfo` filters as instead the core `bloginfo` should be used.

tazotodua:
There was no talk (either by me or whoever) to implement already hooked functions like `_s_*_bloginfo` for `bloginfo`. This is too irrelative to the original purpose we wanted.

However, this is my last word, that you are doing wrong.   Everywhere `_s` is said to be a boilerplate, the start theme, and actually **you have** to be a guide for fellow developers and theme developers, to be skilled and accommodate the standard of theme coding.  Telling that it is not "Genesis theme" and this is lean is just meaningless. That means "lean" theme? Theme is meant to be a theme, a main part of website, and that needs to be as much decent area, as possible, and you should have given the example the people how to code good themes, how to understand that hooks are needed (because there exist plugin developers)..

Every solid bolierplate program has such basic things integrated to guide programmers, but unfortunately you refused to do so  (even more, I can't see any reason what was disadvantages of implementing it, just 1 or 2 line in each .php file), and all themes based on your "boilerplate" are grown up that ugly, as I have to modify tons of themes in order to add navigation menu on custom area (then offered by theme's fixed one location), and many other things. Very pitty.

crunnells:
@tazotodua Thanks for your PR and your input. While you make a good argument for hooks, @grappler is right in that `_s` isn't meant to be _extended_, but _adapted_. Underscores is more of a foundation than a theme system.

We've been holding off on some of the more advanced theme building techniques (eg: SASS only, requiring build tools, adding action hooks, etc) in the interest of keeping the theme easier for people who are just getting into building their own theme. We also use it as the base for most of the themes we build at Automattic. 

That isn't to say that our way is best. When you look at how different themes are built, there are as many methods as there are developers. If you use hooks in all of your themes, that's great! I'd highly encourage you to _fork_ `_s` and make it your own, then share that with your audience rather than abandoning it completely.

Thanks again for your input.

Issue #1328, opened 2018-10-30
Do not override the users alt texts


carolinan:
The theme must respect the users settings.

If the user has added an alt text to their featured images, (or intentionally left it blank) then the theme must show the correct text,  not the title attribute.

samikeijonen:
Is this issue somewhere in `_s` codebase, or general note?

carolinan:
It is in template-tags.php. (I tried to add the code but I'm on mobile at the moment and it just got all messed up. )

carolinan:
Here: https://github.com/Automattic/_s/blob/master/inc/template-tags.php#L137
`<?php
			the_post_thumbnail( 'post-thumbnail', array(
				'alt' => the_title_attribute( array(
					'echo' => false,
				) ),
			) );
			?>`

samikeijonen:
The `<a>` element is already hidden from AT using `aria-hidden="true"` to avoid extra noise. But would be nice to test how img alt inside it works, can it be accessed somehow.

If the link wasn't hidden, I'd still rather use `the_title` because in this case it works as a link text.

grappler:
I agree the title does not always describe the image correctly.

samikeijonen:
It doesn't but if it is used as a link text it's not about image itself anymore.

crunnells:
I'd like to put in a check to see if the image has an `alt` tag, and if not use the title as a fallback (which I think is the spirit of the original code). The other option would be to just use `alt` and `title` fields, regardless if they're blank or not. I'd rather just update the code so it uses the title as a fallback, but figured it's worth discussing other options.

WebAssembler:
Hello, I would like to contribute and make my first pull request on this project. I've cloned the repo and have it set up locally and can see where to edit this code. I agree with @carolinan that the title should either be what is set in the media library or blank as the title is not necessarily helpful. On the Wordpress sample page, for example, the alt text is "Hello world" which is not descriptive to the image I have used. 

If anyone could advise me on the steps for how to do a pull request that would be really useful as I've never contributed to an existing project before. Thank you.

grappler:
@WebAssembler It would be nice if you could create a PR.

The steps that are involved are.
- [Create a fork of this repo](https://help.github.com/articles/fork-a-repo/)
- [Create a new branch](https://help.github.com/articles/creating-and-deleting-branches-within-your-repository/) In this case you could name it something like `1328-thumbnail-alt`
[...]

PR #1348, opened 2019-02-20
Modify post thumbnail output to display the alt text if set.

anthonyhartnell:
When the post thumbnail is output on an index view, the alt text will output what is set in the Media Library for the thumbnail. If it's not set, the alt text will output the post title.

#### Related issue(s):
#1328


WebAssembler:
Would anyone be able to shed some light on what I did wrong? It's my first PR so I'm not really sure what the correct process is. Good learning for next time.

PR #1377, opened 2019-10-29
Image alignment: Make them full-width in mobile devices

### LINKED ISSUE #1376: Can we align `.alignleft` and `.alignright` to be full-width in mobile devices?
The idea is behind a common pattern that, in mobile devices float images often make the text to be unreadable or difficult to read. So, in mobile devices (viewport lesser than `37.5em`) we can make the images full-width to make the text distributed in a nicer way.

## Current
At present the content breaks because of float on mobile devices if the image is small.<br>
floatright

## Expected
Expecting a good-read text top or bottom of the image while in smaller viewport.<br>
floatnone

Actually we did similar in our development. So we thought it'd be nice to have that in core `_s`.

mayeenulislam:
Make `.alignleft` and `.alignright` full-width on mobile devices

Closes #1376

Make `.alignleft` and `.alignright` full-width on mobile devices

### DIFF
--- a/sass/modules/_alignments.scss
+++ b/sass/modules/_alignments.scss
@@ -1,13 +1,26 @@
-.alignleft {
-	display: inline;
-	float: left;
-	margin-right: 1.5em;
+.alignleft,
+.alignright{
+	width: 100%;
+	margin-bottom: 0.5em;
 }
 
-.alignright {
-	display: inline;
-	float: right;
-	margin-left: 1.5em;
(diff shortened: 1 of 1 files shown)


designarti:
That's not a theme issue. This is personalization. Underscores is barebone and it should stay barebone.

philiparthurmoore:
Design decision here - not an actual theme issue that needs to be fixed.

PR #1408, opened 2020-04-16
Use stylelint and eslint directly

neilabraham:

@wordpress/scripts is a huge package with some huge dependencies, but it's only being used for linting, which are just tiny wrapper scripts around stylelint and eslint.  

I propose to use stylelint and eslint directly.  This reduced npm install time from 3 minutes down to 15 seconds on my computer.

#### Related issue(s):

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,9 +17,12 @@
     "url": "https://github.com/Automattic/_s/issues"
   },
   "devDependencies": {
-    "@wordpress/scripts": "^7.2.0",
+    "@wordpress/eslint-plugin": "^4.1.0",
+    "eslint": "^6.8.0",
     "node-sass": "^4.13.1",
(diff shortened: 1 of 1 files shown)


neilabraham:
Hmm, _s fails with the latest version of stylelint-config-wordpress (16.0.0).  @wordpress/scripts uses 13.1.0, but 15.0.0 seems to work here.

Ismail-elkorchi:
Hi @neilabraham thanks for looking at this.

I agree with you that we don't use all the dependencies in @wordpress/scripts, although we may use more of them. The main reason why I used  @wordpress/scripts in the first place is to reduce the number of dependencies that the project relies on as this was one of the main goals that were agreed on in https://github.com/Automattic/_s/issues/1149.

I am inclined to keep this as it is for the moment.

PR #1412, opened 2020-04-22
Added grid-template for no-sidebar

VasusMe:

Modified \sass\layout\_no-sidebar.scss
Added grid-template-columns and grid-template-areas for { .site } when there is no sidebar.

#### Related issue(s):

No Sidebar has the 25% Blank Space. Needs to be adjusted grid-template for no-siderbar templates.

### DIFF
--- a/sass/layout/_no-sidebar.scss
+++ b/sass/layout/_no-sidebar.scss
@@ -1,12 +1,11 @@
 .no-sidebar {
 
-	.content-area {
-		float: none;
-		margin-left: auto;
-		margin-right: auto;
-	}
-
-	.site-main {
-		margin-right: 0;
+	.site {
+		display: grid;
+		grid-template-columns: auto;
+		grid-template-areas:
+			"header"
+			"main"
+			"footer";
 	}
 }


sass/layout/_no-sidebar.scss
```diff
@@ -1,5 +1,15 @@
 .no-sidebar {
 
+	.site {
+		display: grid;
+		grid-template-columns: auto;
+		grid-template-areas:
+			"header"
+			"main"
+			"sidebar"
```
Ismail-elkorchi, line 9: The `.no-sidebar` class is only added when there is no sidebar, so the "sidebar" template area has no effect here and thus it must be removed.

sass/layout/_no-sidebar.scss
```diff
@@ -1,5 +1,15 @@
...
+			"footer";
+	}
+
```
Ismail-elkorchi, line 12: It seems that there is an unexpected whitespace here that must be removed. That's why travis is failing.

Ismail-elkorchi, line 12:
```suggestion

```

sass/layout/_no-sidebar.scss
VasusMe: Right! sidebar needs to remove from grid-template-area.

sass/layout/_no-sidebar.scss
VasusMe: Done

PR #1417, opened 2020-04-26
Add a bundle script to generate a zip file for theme distribution

Ismail-elkorchi:
### Changes proposed in this Pull Request:

This PR adds a NodeJS script to generate a zip file for theme distribution, while excluding development and system files.

### Testing instructions:
To test this PR run `npm run bundle` in command line. A zip file containing only the essential theme files should be created in the parent directory of the actual theme. Note that the name of the zip file name should be the same as the theme directory name.

### DIFF
--- a/bin/bundle.js
+++ b/bin/bundle.js
@@ -13,6 +13,7 @@ const excludes = [
 	'.gitattributes',
 	'.github',
 	'.gitignore',
+	'README.md',
 	'bin',
 	'composer.json',
 	'composer.lock',
--- a/bin/bundle.js
+++ b/bin/bundle.js
@@ -1,0 +1,62 @@
+#!/usr/bin/env node
+
+const path = require( 'path' );
+const fs = require( 'fs' );
+
+const distinationDirectory = 'bundle';
+const excludes = [
+	'.DS_Store',
+	'.stylelintrc.json',
+	'.eslintrc',
+	'.git',
+	'.gitattributes',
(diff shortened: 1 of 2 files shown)


neilabraham:
Would it be better to output a zip file instead of a directory?

Ismail-elkorchi:
> Would it be better to output a zip file instead of a directory?

Definitely, that would be ideal. The problem is that Node.js doesn’t support the full `.zip` format, and I can't find a way to make this work without relying on an external NPM module.

neilabraham:
Having as few npm dependencies as possible was a goal expressed in #1149 in 2017.  Maybe it's time to re-examine that decision.  I'm in favour of adding new dependencies where useful.  I'd quite like to add npm-run-all and nodemon myself.

Ismail-elkorchi:
I updated this PR to output a zip file instead of a directory using [archiver](https://www.npmjs.com/package/archiver). I have tested the command on macOS, and it works without any side effects. Could you please test it on a windows machine, if you have any?

I agree with you that we shouldn't be limited to a fixed number of dependencies if these turn out to be useful. I think that we could add some needed functionality, like SASS watcher, browser-sync and maybe [wp-env](https://make.wordpress.org/core/2020/03/03/wp-env-simple-local-environments-for-wordpress/), while keeping a reasonable number of dependencies.

Feel free to open an issue if you want to discuss this issue further.

PR #1422, opened 2020-05-05
Some simple improvements...

SaidElBakkali:
`ClassName.indexOf` is not a good practice, I think using the DOM API  `node.classList.contains()` is much better.

### DIFF
--- a/js/navigation.js
+++ b/js/navigation.js
@@ -68,7 +68,7 @@
 		while ( self.classList.contains( 'nav-menu' ) ) {
 			// On li elements toggle the class .focus.
 			if ( 'li' === self.tagName.toLowerCase() ) {
-				if ( -1 !== self.classList.contains( 'focus' ) ) {
+				if ( self.classList.contains( 'focus' ) ) {
 					self.className = self.className.replace( ' focus', '' );
 				} else {
 					self.className += ' focus';
--- a/js/navigation.js
+++ b/js/navigation.js
@@ -25,12 +25,12 @@
 		return;
 	}
 
-	if ( -1 === menu.className.indexOf( 'nav-menu' ) ) {
+	if ( menu.classList.contains( 'nav-menu' ) ) {
 		menu.className += ' nav-menu';
 	}
 
 	button.onclick = function() {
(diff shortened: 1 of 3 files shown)


js/navigation.js
```diff
@@ -25,12 +25,12 @@
 		return;
 	}
 
-	if ( -1 === menu.className.indexOf( 'nav-menu' ) ) {
+	if ( menu.classList.contains( 'nav-menu' ) ) {
```
Ismail-elkorchi, line 28: We have to add the `!` (logical NOT) operator here to match the old implementation. We want to check if the `menu` don't contain the `nav-menu` class and not the contrary.

```suggestion
	if ( ! menu.classList.contains( 'nav-menu' ) ) {
```

js/navigation.js
```diff
@@ -65,10 +65,10 @@
 		var self = this;
 
 		// Move up through the ancestors of the current link until we hit .nav-menu.
-		while ( -1 === self.className.indexOf( 'nav-menu' ) ) {
+		while ( self.classList.contains( 'nav-menu' ) ) {
```
Ismail-elkorchi, line 68: The same thing here.

```suggestion
		while ( ! self.classList.contains( 'nav-menu' ) ) {
```

js/navigation.js
```diff
@@ -65,10 +65,10 @@
...
 			// On li elements toggle the class .focus.
 			if ( 'li' === self.tagName.toLowerCase() ) {
-				if ( -1 !== self.className.indexOf( 'focus' ) ) {
+				if ( -1 !== self.classList.contains( 'focus' ) ) {
```
Ismail-elkorchi, line 71: We don't have to test equality since we are using now `contains`.

```suggestion
				if ( self.classList.contains( 'focus' ) ) {
```

js/navigation.js
SaidElBakkali: You're right, I don't know where my head was...

Ismail-elkorchi: it's okay, it happens to the best of us especially in these troubled times.

Ismail-elkorchi:
Congratulations for your first merged PR :tada:

PR #1445, opened 2020-07-10
Use _S_VERSION constant when enqueuing customizer.js

cstielper:
I propose that we use the constant "_S_VERSION" in the wp_enqueue_script function call instead of a static string.

#### Related issue(s):
None.


cstielper:
You're welcome @Ismail-elkorchi 

Happy to see this project being maintained again. 🙂

Issue #1447, opened 2020-07-23
Add backward compatibility for wp_body_open() function


kishanjasani:
The function wp_body_open was added in WordPress 5.2. This theme should call it in header.php
- As lots of the themes in the WordPress repository are submitted with the _s. So backward compatibility is more important.

PR #1448, opened 2020-07-23
Add backward compatibility for wp_body_open function

kishanjasani:
Add backward compatibility for wp_body_open function. as it was introduced in WordPress 5.2. Lots of themes use the _s theme as a base theme. So it's needed to add a check for backward compatibility.
#### Related issue(s): #1447


kishanjasani:
Sorry, Support is already there.

PR #1486, opened 2021-02-26
getting set up

rgothlin:
#### Related issue(s):


kangabell:
This PR looks to be for you personally maybe and not intended for others to use. Is that right @rgothlin? If so, can you fork it to your own account and remove it from this repo?
If not, could you provide a description to explain to others what the purpose of this PR is?

PR #1489, opened 2021-04-16
fix post date bug

iamdavidabayomi:
i choose to use underscore as the based theme for a new client theme and found out that the posted on date outputted two markup date once i change/modified the post.
So a peep into the underscore posted on function i found out that there is a code block outputting both  the published and modified date if the published date is not equal to the modified date.

So i assumed that the proposed behaviour is that if the published date is not equal to the modified date then output the modifed date.

### DIFF
--- a/inc/template-tags.php
+++ b/inc/template-tags.php
@@ -14,15 +14,17 @@ if ( ! function_exists( '_s_posted_on' ) ) :
 	function _s_posted_on() {
 		$time_string = '<time class="entry-date published updated" datetime="%1$s">%2$s</time>';
 		if ( get_the_time( 'U' ) !== get_the_modified_time( 'U' ) ) {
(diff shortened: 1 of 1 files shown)


zimnyd:
It nothing wrong with that. It's a clear info for robots when post was published and when updateed. On front we only need one of them - when the post was updated. To do that we can use of CSS code:
```
// entry-date display only update time
time.published:not(.updated) {
	display: none;
}
```

iamdavidabayomi:
> It nothing wrong with that. It's a clear info for robots when post was published and when updateed. On front we only need one of them - when the post was updated. To do that we can use of CSS code:
> 
> ```
> // entry-date display only update time
> time.published:not(.updated) {
> 	display: none;
> }
> ```

Okay seems i did not check that. Thanks

PR #1497, opened 2021-06-07
Missing Commas in style.css

mav2287:
Added 2 commas that were missing

Addition of 2 missing commas in the stylesheet

#### Related issue(s):


erasmoh:
These commas are not needed here, it will split the target classes, and repeat the main one.

mav2287:
In that case it would appear that both my added commas and the line returns should be removed.

PR #1505, opened 2021-09-24
Update functions.php

wprashed:
We can add post reading time feature

#### Related issue(s):


ravismakwana:
image
Getting error while running npm run watch command. 
I think something is missing, any one have an idea please?

PR #1530, opened 2022-08-19
fix(typo): comment out example function correctly

JakobJR:
typo fix

#### Related issue(s):
none

### DIFF
--- a/inc/custom-header.php
+++ b/inc/custom-header.php
@@ -4,7 +4,7 @@
  *
  * You can add an optional custom header image to header.php like so ...
  *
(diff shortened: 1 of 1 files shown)


adminazhar (commented):
@JakobJR  The best way would be
`*  the_header_image_tag();`

Because php opening tag already exists at top.

JakobJR:
> @JakobJR The best way would be `* the_header_image_tag();`
> 
> Because php opening tag already exists at top.

Yea that's exactly what I did - not sure what you're refering to

PR #1552, opened 2023-12-20
fix: sass update

bracket-web:
I recently got issues with node-sass in node 18 so I updated the sass compiler node-sass to normal sass here is the official sass documentation about that 
https://sass-lang.com/install/ https://prnt.sc/9s5XOenDOOvM

It is also updated .gitignore with yarn.lock because when we use yarn as a package manager, it's not required to load on the main repository. 

#### Related issue(s):

1. Sass update 
2. fixing yarn.lock ignore

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ package-lock.json
 composer.lock
 style.css.map
 .DS_Store
+yarn.lock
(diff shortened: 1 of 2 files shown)


StfBauer:
Works find so far. Thanks for the update.

